PR #186, opened 2026-07-19
fix(backup): handle plasmo-encoded storage values in export and import

Shishir435:
## Summary

Fixes the import failure seen in real backups: **Preferences (Sync)** and **Local Storage** both fail with *"Invalid provider configuration in backup"* while SQLite/Dexie import fine.

**Root cause:** `@plasmohq/storage` persists every value as a JSON string. Backup export reads raw `chrome.storage`, so the provider-config value lands in the backup as an encoded **string** — but `preparePortableStorageImport` validates it against an **array** schema. Every backup that contains provider configs fails the Preferences/Local Storage sections, 100% of the time. Existing tests missed it because they mock `chrome.storage.get` with already-parsed arrays; the real plasmo encoding never appeared in a test.

**Second bug, same cause (export side):** `sanitizePortableValue` can't strip secrets it can't see inside a string. A pre-migration profile whose sync payload still carries `apiKey` would export credentials **verbatim** into the backup zip.

## Fix

- **Import**: accept the plasmo-encoded string form of provider configs — parse before schema validation, still reject malformed JSON. Existing user backups (like the one in the report) now import.
- **Export** (`selectPortableStorageData`): decode plasmo-encoded values → sanitize → re-encode, preserving the stored format byte-for-byte for scalars. Secrets inside encoded provider configs are now stripped.

## Tests

- encoded export round-trip: secret stripped, value stays string-encoded, scalar values byte-identical
- encoded import path: provider config as JSON string imports and persists parsed
- malformed encoded string: still rejected with the validation error
- full `src/lib` suite: 813 passed

## Verification

User-reported repro: v2 backup with provider configs → before: Preferences+Local Storage failed; after: schema validates the parsed array. Chat history (SQLite), vectors, knowledge sets were unaffected before and after.

<h3>Greptile Summary</h3>

This PR fixes backup handling for Plasmo-encoded storage values. The main changes are:

- Decodes values before removing nested secrets during export.
- Accepts encoded provider configurations during import.
- Adds tests for encoded values and malformed provider data.
- Reloads the extension after chat-history import or reset.

This looks safe to merge.

- The updated import path validates decoded provider data.
- Export sanitization now reaches secrets inside encoded values.
- The reload changes use the extension runtime API with a page fallback.
- No blocking issues were found in the changed code.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| src/lib/storage/backup-storage-policy.ts | Decodes encoded storage values before sanitization and restores their encoded form for export. |
| src/lib/backup-service.ts | Parses encoded provider configurations before validating and importing them. |
| src/lib/__tests__/backup-service.test.ts | Adds coverage for encoded export sanitization, encoded imports, and malformed provider data. |
| src/features/knowledge/components/data-migration-settings.tsx | Reloads the extension after a successful chat-history import. |
| src/hooks/use-reset-app-storage.ts | Reloads the extension after resetting chat storage. |

<sub>Reviews (2): Last reviewed commit: ["fix(backup): restart whole extension aft..."](https://github.com/shishir435/ollama-client/commit/e73f99f7f644a884c38d735f225ac8f448b38306) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45435861)</sub>

**Context used:**

- Context used - AGENTS.md ([source](https://app.greptile.com/shishir435/github/Shishir435/ollama-client/-/custom-context?memory=e25238d4-69e0-44c6-b9b5-43440fe11231))

### DIFF
--- a/src/features/knowledge/components/data-migration-settings.tsx
+++ b/src/features/knowledge/components/data-migration-settings.tsx
@@ -20,7 +20,6 @@ import { Button } from "@/components/ui/button"
 import { useConfirmAction } from "@/hooks/use-confirm-action"
 import { useToast } from "@/hooks/use-toast"
 import { backupService, type ImportResult } from "@/lib/backup-service"
-import { MESSAGE_KEYS } from "@/lib/constants/keys"
 import { getDisplayErrorMessage } from "@/lib/error-display"
 import { formatBackupFilenameTimestamp } from "@/lib/format-utils"
 import { logger } from "@/lib/logger"
@@ -132,10 +131,17 @@ export const DataMigrationSettings = () => {
     // Vector/knowledge DB failures are non-fatal for the reload decision.
     const restoredChatHistory = importResult?.database.ok
     if (restoredChatHistory) {
-      browser.runtime
-        .sendMessage({ type: MESSAGE_KEYS.APP.RELOAD })
-        .catch(() => {})
-      window.location.reload()
+      // Restart the whole extension: service worker and every page. The
+      // import bumped the SQLite import generation, so any context that
+      // keeps running with the old generation (background worker, an open
+      // sidepanel) has all of its chat saves silently skipped by the
+      // stale-writer guard. A page-level reload fan-out cannot guarantee
+      // delivery — runtime.reload() can.
+      try {
+        browser.runtime.reload()
+      } catch {
+        window.location.reload()
+      }
     }
   }
 
--- a/src/hooks/use-reset-app-storage.ts
+++ b/src/hooks/use-reset-app-storage.ts
@@ -1,5 +1,5 @@
 import { useCallback } from "react"
-import { MESSAGE_KEYS } from "@/lib/constants"
+import { browser } from "wxt/browser"
 import { feedbackService } from "@/lib/embeddings/feedback-service"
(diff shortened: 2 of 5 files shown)


Shishir435:
Second bug from the same import session fixed in `e73f99f7`: **new chats silently stop persisting after a backup import.**

**Root cause:** the import bumps the SQLite import-generation marker (correct — stops stale contexts clobbering the restored DB). But the "Reload Extension" button only ran `window.location.reload()` on the options page plus a fire-and-forget `APP.RELOAD` message. Two holes:

1. The **background service worker has no `APP.RELOAD` listener at all** — it keeps its pre-import generation until it happens to idle out.
2. An open **sidepanel can miss the message** because the options page reloads mid-send.

Any context left on the old generation has every SQLite save skipped by the stale-writer guard (`Skipping SQLite save from stale context after backup restore` in the console) — chats live in memory, vanish on next load.

**Fix:** import and reset flows now call `browser.runtime.reload()` — restarts the SW and every page atomically, `location.reload()` fallback. The reset flow had the same fan-out and additionally cleared the generation key it had just bumped.

**Immediate workaround for anyone hit by this:** toggle the extension off/on in `chrome://extensions` (or restart the browser) — any chat created after a full restart persists fine. Chats created in the stale window are unrecoverable (they were never written).

PR #207, opened 2026-07-26
fix(errors): report fidelity follow-up for 0.12.4

Shishir435:
Follow-up to #206. Four defects found reviewing the 0.12.4 error-reporting work end to end, plus the release notes it left behind. All are report-quality or cost, none block the release.

## Cost: one self-test run instead of one per bubble

`getBundle` ran the full self-test suite unconditionally, and the error panel prepares a bundle on mount. The message list is virtualized, so a screenful of failures meant one repository transaction, two storage round-trips, one provider network probe, and one recorded diagnostic event **per visible failure** — the displayed errors diluted the ring buffer that exists to explain them.

Results are now shared for 60s with in-flight join, and only a real execution records an event.

Cancellation is preserved rather than traded away: the shared run owns its own `AbortController`, an aborting caller stops waiting immediately, and the underlying work is cancelled only once every interested caller has abandoned it. One client's RPC timeout can no longer kill another client's in-flight suite. `diagnostics.run` passes `force` — it is only reachable from pressing **Run self-tests**, which means measure now.

## Reports no longer contradict themselves

`checks.providerEnabled` was inferred from a successful `listModels`. That is a different claim: discovery runs with `enabledOnly: false`, so it succeeds against a *disabled* provider. Real reports asserted `Provider enabled: yes` directly above a diagnostics line reading `disabled`. Now read from provider config via `providers.list`.

Failure-path latency is kept — an instant refusal and a 30s timeout are different bugs — but labelled `(time to failure)` so it cannot be misread beside `reachable: no`.

## One draft composer, one entry point

There were **four** issue-draft builders: the error panel, the footer's bare duplicate for the same message, the generic help draft, and a hand-rolled one in `diagnostics-settings.tsx` with no diagnostics paste block, no privacy statement, and no URL-length clamp — missing in the one place a reporter arrives already holding a bundle.

All now share `REPORT_TAIL` and a `composeReportUrl` that keeps the length guarantee. `buildDiagnosticIssueUrl` moves into `error-report.ts` with its fields clamped, and its test moves with it.

The footer's link is deleted. It rendered for exactly the messages that already show the panel, two rows below it, with no bundle and no local checks — the more discoverable button produced the weaker report.

## Stale recovery actions expire

A week-old failure still offered a button that would flip a provider setting and drop a fresh response into the middle of that conversation. Actions that mutate config or append a turn now declare `maxAgeMs` (1h); read-only navigation declares none. Because resolution is first-match-wins, a stale in-place fix degrades to the settings deep link rather than to nothing.

## Release notes

The 0.12.4 entry was written in #202, before #204 and #205/#206 landed, so the uninstall flow and the entire error-reporting feature were absent from it while sitting under `Unreleased`. Folded into one 0.12.4 entry and condensed to summary bullets.

Process note: a release-notes commit that lands mid-line cannot describe the rest of the line. Worth a final pass at tag time.

## Verification

- 271 test files / 2,220 tests pass
- Biome check, format check, typecheck clean
- New coverage: shared-run dedupe and abort isolation, config-read `providerEnabled`, stale vs fresh recovery resolution, labelled failure latency, diagnostics draft carrying the shared tail and staying inside the URL limit

<h3>Greptile Summary</h3>

This follow-up improves diagnostic accuracy, report consistency, and stale-action handling.
- Forced self-test requests now start a fresh execution while automatic bundle requests share recent or in-flight results.
- Diagnostic cache publication is sequence-guarded so older executions cannot replace newer measurements.
- Provider state and failure latency are reported more accurately.
- Issue-report entry points now share privacy, diagnostics, and URL-length safeguards.
- Recovery actions that mutate state or append turns expire after one hour.
- Release notes consolidate the completed 0.12.4 work.

The PR appears safe to merge.

No blocking failure remains; the forced diagnostics path starts its own execution, and sequence ordering prevents an older overlapping run from replacing its results.

| Filename | Overview |
|----------|----------|
| src/lib/diagnostics/diagnostics-service.ts | Adds TTL and in-flight sharing, isolated cancellation, unconditional fresh forced executions, and sequence-guarded cache publication. |
| src/background/rpc-server.ts | Routes the user-triggered diagnostics run through the forced execution path. |
| src/lib/error-report.ts | Consolidates report composition and adds diagnostic draft clamping and clearer failure-latency labeling. |
| src/features/chat/lib/error-recovery-actions.ts | Expires state-changing recovery actions while preserving read-only settings navigation. |
| src/features/chat/components/chat-error-report-action.tsx | Reads provider enabled state from configuration and supplies message age to recovery resolution. |

</details>

```mermaid
sequenceDiagram
  participant U as User
  participant RPC as Diagnostics RPC
  participant S as DiagnosticsService
  participant A as Automatic bundle run
  participant F as Forced self-test run
  U->>RPC: Run self-tests
  RPC->>S: "run(signal, force=true)"
  S->>F: Start new sequenced execution
  Note over S,A: Existing automatic execution may continue
  F-->>S: Publish only if sequence is newest
  S-->>RPC: Fresh test results
  RPC-->>U: Display results
  A-->>S: Older result ignored for cache publication
```
</details>

<sub>Reviews (4): Last reviewed commit: ["docs(diagnostics): record why forced run..."](https://github.com/shishir435/ollama-client/commit/c8df2f7fa6d8b50b8c19dc008551f3d234159f5a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47361341)</sub>


Shishir435:
Fixed in 12fe6637 — the P1 was real. `force` skipped the completed-result cache but then fell through to `joinSharedRun(sharedRun)`, so pressing **Run self-tests** during an in-flight bundle request returned pre-click state as current state. Exactly the case the flag exists to prevent.

A forced call now always starts its own execution via `startSharedRun()`. It still becomes the run later callers attach to, so forcing does not multiply concurrent suites.

One consequence the review did not name: a forced run can now overtake an older one, and the older one would otherwise publish its stale results to the cache when it finished. Cache writes are therefore ordered by a per-run sequence number — a late finisher cannot overwrite newer results.

Regression test covers the ordering explicitly: start an automatic run, force a second, resolve the *newer* one first, then the older, and assert the cached result is the newer measurement.

271 files / 2,221 tests, Biome, typecheck clean.

Shishir435:
Closed in f7c4d4ff.

A forced caller now joins an in-flight run whose `startedAt` is at or after its own request time. Such a run is by definition measuring the state the caller is asking about, so sharing it costs no freshness — a double click or two open options pages get one suite and one completion event. A run that began *before* the request is still never joined, which is the P1 from the previous round.

Worth noting what this exchange surfaced: the earlier regression test started an automatic run and a forced run in the same millisecond, which cannot represent a human clicking after a request began. It now advances a faked clock by 2s between the two, so it asserts the actual condition — the gap is what makes an in-flight measurement potentially stale — rather than same-tick ordering.

Remaining honest limit: staleness is inferred from time, not from tracked config mutations. A change committed in the same millisecond a run starts would not be detected. Given a mutation requires user interaction, that window is not reachable in practice; tracking a config-mutation counter would be the stricter fix if it ever matters.

271 files / 2,222 tests, Biome, typecheck clean.

Shishir435:
Correct, and the right conclusion is to drop the coalescing rather than patch the comparison — reverted in 6be0b166, rationale recorded in c8df2f7f.

The clock comparison cannot be made exact. Millisecond granularity makes a run that began just before the request indistinguishable from one that began just after, and freshness does not actually depend on start order — it depends on whether configuration changed in between, which nothing here tracks. Tightening `>=` to `>` would only move the ambiguity, not remove it.

So `force` is back to always starting its own execution, which was the state you graded 4/5 last round. Reassessing what the coalescing was buying: the only caller that can force is the **Run self-tests** button, and it is `disabled={running}`, so a double click cannot produce two forced runs. The remaining case is two open options pages measuring at different times — which *should* measure twice. The saving was speculative; the ambiguity it introduced was not.

Left a note at the branch point so this is not reattempted with a clock. If it ever needs coalescing, the workable shape is a config-mutation counter: join an in-flight run only when no mutation has landed since it started. That is exact, and cheap to add if a real duplicate-suite problem shows up.

Automatic runs are unaffected — bundle requests still share one suite for 60s, which is where the actual cost was.

271 files / 2,221 tests, Biome, typecheck clean.

PR #220, opened 2026-07-29
0.12.6: provider sizes, unexplained turn failures, transcript correctness, CI browser gates

Shishir435:
Eight commits continuing the `0.12.6` cycle: two user-visible failures that had no
explanation, the remaining English on the send path, YouTube transcript
correctness, and the browser gates finally running in CI.

## Provider metadata

`/v1/models` reports no size — the OpenAI schema has no field for one, and
neither vLLM, LocalAI, KoboldCPP nor an LM Studio fallback list adds one. A
self-hosted `Qwen3-8B` therefore showed a blank parameter badge beside genuine
sizes from Ollama and llama.cpp. This reuses the LM Studio id parser rather than
adding a second convention, so ids that state no size unambiguously
(`gpt-4o-mini`) stay blank instead of having one invented from a version number.

## A turn that finishes with no answer

A provider can close the stream as a clean success with no content, no thinking,
and no tool run — a model unloaded mid-turn, or a conversation with no room left
to reply. That finalized as an empty assistant bubble: nothing to read, nothing
to report, no retry. It is the shape behind "connected, model loaded, says
nothing back".

The reducer now reports *why* a finished turn is empty instead of composing copy,
which also moves the pre-existing thinking-only fallback out of a pure module
where it could never be translated. A user-initiated stop is excluded — that
finalizes through the abort path, which owns its own wording.
`metrics.emptyResponse` gives the turn the same retry an interrupted answer gets.

## The send path speaks the chosen language

Eight strings were English regardless of locale: the precondition toasts, the
session-create and message-save failures, the context-preparation failure (its
toast *and* the assistant message it persists), the grounded-only refusal, the
context-trimmed notice, and the retrieval warning the background sends over the
port.

Two of those live in modules that deliberately have no `t` — `turn-preparation`
is pure, and context building runs in the background — so `TurnToast` now carries
a message key plus interpolation values and the extension page resolves it. That
is also what stops the next warning added there from being English again. The
grounded-only message links the setting by its own label key, so renaming the
setting cannot leave the message pointing at something the user cannot find.

## YouTube transcripts come from the video you are watching

`ytInitialPlayerResponse` is an inline script from whichever video the document
first loaded, and YouTube navigates without a reload. Move to a second video and
that script still describes the first one, caption `baseUrl` included — so a
question about the open video could be answered from a different video's
transcript with nothing to indicate it.

The payload is now trusted only when it names the video in the address bar;
otherwise the watch document is refetched same-origin, which carries the user's
session the way the player's own request does. A matching id costs no extra
request. Caption choice also stops assuming English: author-written captions in
the browser's language first, English still the fallback ahead of speech
recognition. Timestamps stay on every line — they are what makes a transcript
answerable about *when* something was said.

## Historical storage, decided rather than left open

Profiles whose chat history predates the SQLite backend (written by `0.6.3` or
earlier, never migrated because that bridge shipped only in `0.6.5`–`0.7.3` and
ran on side-panel mount) initialize clean by decision. Two things make that
honest:

- a test pins the fresh-init path. The dangerous failure is not "no data" but a
  *rejecting* migration, which would leave the marker on `legacy`, retry every
  boot, keep the sql.js path live, and land writes where the OPFS owner is not
  reading.
- backups written by `0.6.3` or earlier carry chats as `chat-db.json`, a Dexie
  export. Import reported "Missing database.sqlite", which reads as a corrupt
  archive rather than an old one. It now says which backup it is; settings,
  embeddings and knowledge sets still restore.

Import results gained an `errorKey` alongside `error` for this, since the service
runs where there is no `t`: exception text stays raw, a recognized state the user
can act on gets translated copy.

## CI runs the browser gates

The Chromium gates covering production boot, OPFS restart durability and sql.js
migration idempotence have existed since `0.12.5` and only ever ran by hand.
Wired in without inflating the run:

- `needs: [checks, build]`, so the most expensive minutes are never spent on a
  branch failing typecheck or a unit test
- no extra builds — the chrome leg of `build` also produces the benchmark target
  and uploads both outputs; the gates boot the exact bytes CI packaged
- Chromium only, browser download cached on the Playwright version
- draft PRs skipped, `skip-e2e` label escape hatch, failure artifacts only
- `workflow_dispatch` added, since CI otherwise triggers on pull requests and
  pushes to main only

The eight commits in this PR have had no CI: they landed after the previous PR
merged, and a push to a feature branch with no open PR triggers nothing. Local
hooks (lint-staged, then `test:run` on push) are all that has run against them.

## Verification

`pnpm typecheck`, `pnpm lint:check`, `pnpm test:run` (291 files, 2396 tests) and
`pnpm generate:resources` all clean locally. The three `@critical` browser specs
have not been run against these commits — that is what this PR is for.

<h3>Greptile Summary</h3>

The PR completes the reported follow-up fixes and expands the 0.12.6 release work.
- Validates YouTube transcript sources against the current video and safely handles mounted-panel fallbacks.
- Pins CI actions to immutable commits, limits workflow permissions, and adds critical Chromium browser gates.
- Adds localized chat failure messages, retryable empty-response handling, provider-size metadata, and clearer legacy-backup results.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/transcript-extractor.ts | Validates inline caption metadata and mounted transcript panels against the current YouTube video; the previously reported stale-source paths are addressed. |
| .github/workflows/ci.yml | Pins workflow actions, declares read-only repository permissions, and runs critical browser tests against packaged Chromium builds. |
| src/features/chat/hooks/chat-stream-reducer.ts | Classifies thinking-only and entirely empty successful streams so callers can render localized explanations and offer retries. |
| src/features/chat/hooks/use-chat-turn-controller.ts | Resolves send-path notices through translation keys and persists localized assistant failure messages. |
| src/lib/backup-service.ts | Distinguishes unsupported legacy chat exports from malformed backups while preserving structured translation keys. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(youtube): tie the panel fallback to ..."](https://github.com/shishir435/ollama-client/commit/37e969f5ebd016eda3cd72a5ca11080e1eab7f6b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48295655)</sub>


Shishir435:
Correct again, and the pattern in my last two attempts is worth naming: I kept moving the unverifiable read rather than making it verifiable. Fixed properly in `37e969f5`.

**The panel has no id — but the document does.** The same staleness that makes the inline player response untrustworthy as *data* makes it decisive as *identity*. It is part of the document, so:

- inline payload still names the video in the address bar → no navigation since load → everything mounted, transcript panel included, was rendered for this video
- it names a different video, or names none → the document has moved on, and its unlabelled DOM cannot be tied to anything

```ts
const documentBelongsToCurrentVideo = (): boolean => {
  const videoId = currentYouTubeVideoId()
  if (!videoId) return false
  return getYouTubePlayerResponse()?.videoDetails?.videoId === videoId
}
```

So the fallback now splits by provenance:

| panel state | decision |
| --- | --- |
| already mounted, document names current video | read it |
| already mounted, document stale or unnamed | **return nothing** |
| not mounted, we open it | read it — whatever the click renders is built by the live page for the video it is showing, current by construction |

Returning nothing in the middle row is deliberate: no transcript is a true and checkable answer, a previous video's transcript is neither.

**No new selectors, on purpose.** I considered reading a video id from `ytd-watch-flexy`, `#movie_player`, or the canonical link. All are internal names YouTube changes without notice, so a wrong guess would add a second source of silent wrongness rather than a check on the first. The document's own payload is already parsed and already trusted for exactly this question.

**Test fixtures:** five of them mounted a transcript panel with no inline player response at all — a shape no real watch page has, and the reason they passed before. They now describe a loaded document whose caption fetch failed, which is the situation the fallback exists for. Two new tests cover this directly: a stale document with a mounted panel returns null, and a freshly-clicked panel is still read when the document is stale. Both verified to fail with the gate removed.

Suite: 2402 tests, typecheck and lint clean.

PR #228, opened 2026-07-30
i18n(chat): translate activity trace labels from persisted keys

Shishir435:
## Why

The activity trace persisted its English label with each event, so a turn built on one device rendered that language forever — switching locale left chat history untranslated. `AGENTS.md` §5 lists this as a rolling i18n tail: services return message keys, presentation translates.

## What

- `ActivityEvent.labelKey` and `UsedContextChunk.titleKey` are persisted beside their text. Services emit the key; the trace and sources sheet translate it and fall back to the stored text for turns written before the key existed.
- Keys live in `src/application/context/activity-labels.ts`. The key is per event, not derived from `kind`, so one kind can carry distinct steps — same distinction `ToolRun.displayNameKey` already keeps for tools.
- `"Selected tab context"` (tab fallback chunk) is now `chat.sources.tab_context`.
- Result counts use a `results_one` / `results_other` plural pair instead of hand-built `"1 result"` / `"2 results"`.
- Dropped `getDisplayLabel`, an identity wrapper in the file, now that labels arrive translated.
- New keys added to all nine locales; `pnpm generate:resources` re-run (no `_locales` diff — the `extension` block is untouched).

## Note for review

`ActivityEvent` and `UsedContextChunk` are each declared twice — once in `src/types/chat.schemas.ts` for persistence, once in `src/protocol/streams/chat-stream.ts` for the stream. Zod strips unknown keys, so a field added to one and not the other vanishes silently between background and sidepanel. Both copies carry the new fields, and both paths now have a contract test that fails if a field is dropped. Collapsing the duplication is out of scope here and recorded in the roadmap.

## Still open

Activity `outputPreview` / `sourceTitles` prose ("Recalled past conversation context", "Previous conversation") stays English. Those previews mix generated prose with retrieved content, so they need their own shape rather than a second label key.

## Gates

- `pnpm typecheck` ✅
- `pnpm test:run` ✅ 304 files, 2503 tests
- `pnpm generate:resources` ✅

<h3>Greptile Summary</h3>

Adds persisted translation keys for activity events and context-chunk titles, translates them at presentation time with legacy-text fallback, and completes Russian CLDR plural forms for activity result counts.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/application/context/activity-labels.ts | Centralizes application-owned activity and context-chunk translation keys with English persistence fallbacks. |
| src/application/context/build-context.ts | Emits translation keys alongside persisted activity labels and the generated tab-context title. |
| src/features/chat/components/reasoning-trace-events.tsx | Resolves persisted activity keys through i18next and formats result counts using locale-aware pluralization. |
| src/features/chat/components/reasoning-trace.tsx | Threads the translator through compact and expanded activity-trace rendering. |
| src/i18n/__tests__/locale-architecture.test.ts | Adds CLDR plural-category catalog validation and direct Russian result-count resolution coverage. |
| src/locales/ru/translation.json | Adds the required one, few, many, and other forms for activity results, resolving the previously reported Russian fallback. |
| src/protocol/streams/chat-stream.ts | Preserves the new optional translation-key fields across the stream protocol. |
| src/types/chat.schemas.ts | Persists optional activity-label and context-title keys while remaining compatible with older stored turns. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(i18n): complete Russian plural categ..."](https://github.com/shishir435/ollama-client/commit/ea7ba48f8f94cb1773ce9ed79e6e59853b306018) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48760262)</sub>


Shishir435:
Fixed in ea7ba48f — the review call was correct, and the failure is worse than a raw key: it silently renders the English string.

Verified rather than assumed. `Intl.PluralRules('ru')` selects `few` for 2–4 and `many` for 0 and 5–20, so with only `_one`/`_other` present a real i18next instance resolved `results` at count 2 to `"2 results"` — English text inside a Russian UI, no missing-key warning.

- `chat.reasoning.trace.results` now has `_one` / `_few` / `_many` / `_other`.
- Same gap existed in `settings.loaded_models.count` (pre-existing, not from this PR) — fixed alongside, since it is the same defect in the same locale.
- Only ru needed it. `es` / `fr` / `it` declare a `many` category, but it is unreachable below a million, so result counts never select it.

Two guards, because key-presence alone would not catch a wrong suffix:

1. `locale-architecture` asserts every locale covers each plural category its own counts can reach, derived from `Intl.PluralRules` over 0–500 rather than the full category list. Scoped to families that opt into plurals by declaring `_one`, so deliberate single-form `{{count}}` keys stay untouched.
2. One case resolves Russian counts through a real i18next instance and asserts the rendered strings.

Both were confirmed to fail on the original catalog before the fix landed.

The `reasoning-trace` test mock faked plurals with `count === 1` — that assumption is what hid this — so it now selects the category the way i18next does.

Gates: `typecheck`, `lint:check`, `format:check`, `generate:resources`, `test:run` (304 files, 2505 tests) all green.

Unrelated find, not touched here: `chat.sources.rag_source_plural` and `view_retrieved_plural` use the i18next v3 `_plural` suffix. With JSON v4 those keys never resolve. Separate change.

PR #229, opened 2026-07-30
Verify chat-history migration per table in 0.12.6

Shishir435:
Backports #227 to the 0.12.6 line and makes its evidence readable, so the evidence clock for retiring the sql.js fallback starts with the release shipping tomorrow instead of with 0.13.0.

## Why into 0.12.6 rather than waiting

The removal gate in `compatibility-ledger.json` requires the fallback to "complete its evidence window" — 2–3 months of clean field receipts. That window cannot start until the code that writes receipts reaches real profiles. Landing it here rather than in 0.13.0 moves the earliest possible removal forward by roughly the length of the 0.12.6 → 0.13.0 gap, at no cost to the 0.13.x plan.

## Commits

1. **`feat(persistence): verify chat-history migration per table`** — cherry-pick of #227, applied clean with no conflicts.
2. **`feat(diagnostics): make migration evidence readable`** — new here, not in #227.
3. **`docs(changelog): record migration verification in 0.12.6`**

## The table-list difference from #227

`DURABLE_TABLES` names the **seven** tables 0.12.6's schema creates, not the ten on `release/0.13.x`. `turn_runs`, `ingestion_runs`, and `model_pull_runs` are all 0.13.x tables, so a legacy blob from 0.6.0–0.12.2 cannot contain rows in them — they are created empty after import either way. Verifying seven tables here verifies exactly as much real history as verifying ten does.

The contract test derives the expected list from `schema.ts` plus its migrations, so this cannot drift: adding a table without adding it to `DURABLE_TABLES` fails the test. Two fixtures that used a 0.13-only table as their "destination gained an empty table" example now use `chunk_feedback`, which 0.12.6 has.

## Why commit 2 exists

Without it, this release starts a clock nobody can read. `writeMigrationReceipt` wrote to device-local storage and **nothing outside `src/lib/persistence` read it**. A profile could fail verification, get served from the retained blob, and the only outward signal was `Storage backend: legacy` in a report — which reads identically to a migration that never started. Three months of that produces silence, not evidence.

The diagnostics bundle now carries a migration summary and the support draft prints it: outcome, attempts, source schema version, integrity verdicts, orphan rows, per-table shortfalls, failure text.

**Row counts do not leave the device.** `sessions: 1482` describes how much someone has said. The summary keeps only tables that arrived *short*, as `table:source→imported`, capped at ten; integrity output is clamped since a damaged file can produce a long report. A test asserts the counts are absent from the serialized summary. A clean migration is one line; a profile that never had a legacy blob reports nothing.

## Ledger

`persistence-legacy-override-switch` is recorded as `introducedIn: 0.12.6` rather than `0.13.0`, since that is now where it ships.

## Gates

- `pnpm typecheck` ✅
- `pnpm test:run` ✅ 294 files, 2442 tests
- `pnpm generate:resources` ✅

**Needs before release:** the Critical Chromium browser gates on this exact commit, per release rule 2 — this PR brings #227's CI workflow additions with it, so those gates run here for the first time on the 0.12.6 line.

## Not in this PR

The engine unification (sqlite-wasm replacing sql.js for reading the legacy blob) stays out of a release shipping tomorrow — it rewrites the live path for exactly the profiles already in a failure state, and it needs its own soak. That is 0.13.x work, and it does not need the evidence window.

<h3>Greptile Summary</h3>

This PR strengthens legacy chat-history migration verification and exposes privacy-reduced migration evidence in diagnostics.

- Verifies row counts across all seven durable 0.12.6 tables and rejects imported databases missing required tables.
- Repairs missing tables when opening existing databases and adds rollback protection around database replacement.
- Records migration outcomes and adds a redacted summary to diagnostic bundles and support reports.
- Adds packaged Chromium and opt-in Firefox migration gates with retained evidence.
- Updates the compatibility ledger and 0.12.6 changelog.

The PR appears safe to merge.

No blocking failure remains; the required-table verification now repairs and checks missing durable tables, while diagnostic migration summaries expose only count deltas rather than absolute history volumes.

| Filename | Overview |
|----------|----------|
| src/lib/persistence/owner-host.ts | Extends migration verification to compare every source table, reject missing destination tables, and record migration receipts. |
| src/lib/persistence/chat-db-worker.ts | Initializes missing schema objects on existing databases and protects database imports with staged rollback and integrity verification. |
| src/lib/persistence/durable-tables.ts | Defines the 0.12.6 durable-table inventory and centralizes count, presence, integrity, and mismatch checks. |
| src/lib/persistence/import-rollback.ts | Adds rollback staging and interrupted-import recovery for OPFS database replacement. |
| src/lib/diagnostics/diagnostics-service.ts | Reduces local migration receipts to bounded diagnostic evidence without exposing absolute source or imported row counts. |
| src/lib/error-report.ts | Adds concise migration outcomes and failure details to user-initiated support reports. |
| .github/workflows/ci.yml | Adds packaged Chromium migration verification and an opt-in Firefox migration workflow with retained evidence. |

</details>

```mermaid
sequenceDiagram
  participant Host as Persistence owner
  participant Legacy as Legacy sql.js blob
  participant Worker as OPFS SQLite worker
  participant Receipt as Local migration receipt
  participant Diagnostics as Diagnostics bundle

  Host->>Legacy: Read blob, table counts, and integrity
  Host->>Worker: Import database bytes
  Worker->>Worker: Stage rollback copy
  Worker->>Worker: Replace and reopen database
  Worker->>Worker: Create missing schema objects
  Worker-->>Host: Imported table counts and integrity
  Host->>Host: Compare every durable table
  alt Verification succeeds
    Host->>Receipt: Record migrated outcome
    Host->>Host: Select OPFS backend
  else Verification fails
    Worker->>Worker: Restore rollback copy
    Host->>Receipt: Record failure evidence
    Host->>Host: Retain legacy backend
  end
  Diagnostics->>Receipt: Read device-local receipt
  Diagnostics->>Diagnostics: Remove absolute row counts
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(persistence): apply the schema on ev..."](https://github.com/shishir435/ollama-client/commit/63c277bf3860d52e774eec655c19f80d54f85f92) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48783370)</sub>

**Context used:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,116 @@ jobs:
           retention-days: 1
           if-no-files-found: error
 
+  persistence-migration:
+    name: Persistence migration gates (chrome)
+    # The evidence gate for retiring the legacy sql.js backend: a packaged
+    # Chromium profile migrates a real legacy blob, is interrupted mid-flight,
+    # and is verified table by table. The report is uploaded on success too —
+    # a gate whose evidence only exists on a laptop cannot be reviewed.
+    needs: [checks, build]
+    if: >-
+      github.event_name != 'pull_request' ||
+      (github.event.pull_request.draft == false &&
+       !contains(github.event.pull_request.labels.*.name, 'skip-e2e'))
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version: 24
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Download packaged builds
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: chrome-e2e-builds
+          path: build
+
+      - name: Resolve Playwright version
+        id: playwright
+        run: |
+          echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ steps.playwright.outputs.version }}
+
+      - name: Install Chromium
+        run: pnpm exec playwright install --with-deps chromium
+
+      # Under xvfb so the runner's own headless→headful retry has a display to
+      # fall back to when extension-id resolution fails headless.
+      - name: Verify OPFS migration
+        run: xvfb-run --auto-servernum pnpm verify:opfs-migration
+
+      - name: Upload migration evidence
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: persistence-migration-evidence-chrome
+          path: artifacts/persistence-benchmark
+          retention-days: 30
+          if-no-files-found: error
+
+  persistence-migration-firefox:
+    name: Persistence migration gates (firefox)
+    # Opt-in: the Firefox runner drives geckodriver against a temporary add-on
+    # and a pinned extension UUID, which is slower and more fragile than the
+    # Chromium path. Dispatch it before a release, or when persistence changes.
+    needs: [checks, build]
+    if: github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version: 24
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      # Downloaded from Mozilla rather than installed through a third-party
+      # action or the snap package: the snap build cannot use an arbitrary
+      # profile directory, which the restart scenario depends on.
+      - name: Install Firefox
+        run: |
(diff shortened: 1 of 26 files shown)


Shishir435:
Both findings were real. Fixed in 63c277bf.

## Missing tables escaped verification — confirmed, with one correction

The mechanism is exactly as described: `initializeSchema` ran `SCHEMA_SQL` only when `sessions` was absent, so a table added to the schema without a paired migration never reached a profile that already had chats. `repairSchemaDrift` covers only `tool_loop_runs` and `prompt_templates`.

Verified which tables can actually hit this:

| Table | Entered schema | Migration? | Reachable |
|---|---|---|---|
| `chunk_feedback` | 0.10.0 (`70e1ecc8`) | none | **yes** |
| `files` | first SQLite commit (`6442276f`) | n/a | no |
| `kv_store` | first SQLite commit (`6442276f`) | n/a | no |

So `files` and `kv_store` cannot be missing from a database that has `sessions` — they shipped together. `chunk_feedback` is the live case, and it predates this PR: profiles created on 0.6.0–0.9.x have been missing it since 0.10.0. What this PR did wrong was report those migrations as verified.

**Fix.** All 16 statements in `SCHEMA_SQL` are `IF NOT EXISTS`, so it now runs on every open and fills in whatever is absent. The version stamp stays gated on a fresh database — stamping an old one would skip forward migrations it still needs.

**Plus a guard**, because the root cause was a class of defect rather than one table: verification now also checks the *destination's* table list via `findMissingDurableTables`. A count comparison structurally cannot catch this (an absent source table is skipped on the assumption a migration creates it), so the next table added without a migration fails the migration instead of shipping a profile that raises "no such table" at its first write.

Also changelogged under Fixed — those profiles gain the table on next open, independent of any migration.

## Row counts leaving the device — confirmed, and my test asserted the contradiction

Correct, and worse than described: the counts travelled by **two** routes.

1. The summary formatted `table:source→imported`, which `migrationLines` copied into the issue body.
2. `describeMismatches` formatted `table imported/source` into the thrown error, which becomes the receipt's `failure` — also printed in the report.

The test I wrote asserted `mismatches: ["messages:39204→39199"]` as expected output while claiming counts stay local, and its privacy assertion only checked the *sessions* count (`1482`), which was never in the mismatch string. It passed while both message counts left.

**Fix.** Both formatters emit `table short by N`. The shortfall is the diagnosable half — "messages lost 5 rows" is what a maintainer acts on; `39204` only says how much someone has said. Absolute counts stay in the receipt's structured `sourceCounts`/`importedCounts`, which never leave the device.

The test now loops over every count in the fixture (`1482`, `39204`, `39199`) and asserts none appears in the serialized summary, and the report-body test asserts the same for the draft.

## Gates

`typecheck`, `lint:check`, `format:check` ✅ · `test:run` ✅ 294 files, 2443 tests

Migration fixtures updated: they had listed partial table maps, which a real imported database cannot be now that the schema is applied on every open.

PR #230, opened 2026-07-30
Survey the legacy blob on sqlite-wasm

Shishir435:
Step one of the engine unification: the migration no longer needs sql.js to read its source.

## The problem

The migration measured its source blob with sql.js (`countLegacyRows` → `initSqlJs` → `fetch(assets/sql-wasm.wasm)`) and imported it with sqlite-wasm. Two SQLite builds read the same file, and 644KB of wasm shipped for a read the engine already running in the worker performs natively.

Worse for the evidence window: verification compared what sql.js saw against what sqlite-wasm produced. A disagreement between engines would surface as a table-count mismatch — indistinguishable from a real migration defect, and the failure mode most likely to make a clean window look dirty.

## The change

A read-only `surveyDb` op on the persistence worker: import the candidate to the scratch path the import path already uses, count every durable table, read `PRAGMA user_version`, take the integrity verdicts, then close and unlink. Nothing writes to that handle — a source blob is evidence, and evidence that has been written to is no longer what was measured.

Verification now compares one engine against itself.

`legacy-blob-reader.ts` keeps only the IndexedDB read, which needs no engine at all: 96 lines → 41.

## What still uses sql.js

In production, exactly one module: `lib/sqlite/legacy-db.ts`, the live fallback that serves history to a profile still on the blob. A contract test in `architecture-boundaries.test.ts` pins that and fails on any new runtime import.

Three dev-only entrypoints are allowlisted (`config/wxt-hooks.ts` strips them from store builds). One of them — the migration-verification harness — legitimately needs sql.js, because its fixtures have to *write* blobs in the old topology. That is all sql.js was ever needed for, per §4C.

## What this does and does not unblock

- **Clears** the §4B gate "official SQLite-WASM owner replaces `countLegacyRows`". No official-owner work now stands between here and removal.
- **Does not** remove the 644KB asset. That waits on the live fallback, which is gated on the evidence window — a data decision, not an engine one.

Removal day is now: delete `legacy-db.ts`, flip a failed marker read from "answer legacy" to an explicit failure, drop the dep and the asset. No engine work left in it.

## Review note

Test harnesses that stubbed `countLegacyRows` on the reader module now queue `surveyDb` answers on the stub worker instead, since the survey crossed into the worker. Same scenarios, same assertions.

## Gates

- `pnpm typecheck` ✅
- `pnpm test:run` ✅ 304 files, 2506 tests

The Chromium critical persistence gates exercise the real migration path against a packaged build, so they are the meaningful check on this one — unit tests stub the worker by construction.

<h3>Greptile Summary</h3>

The PR moves legacy-blob surveying from sql.js into the sqlite-wasm persistence worker.
- Adds a read-only `surveyDb` worker operation that reports durable-table counts, schema version, and integrity results.
- Ensures scratch databases are closed and unlinked across import, open, survey, and cleanup failures.
- Removes sql.js-based surveying from the legacy blob reader and updates migration verification tests.
- Adds an architecture boundary test limiting production sql.js usage to the legacy fallback.

The PR appears safe to merge.

No blocking failure remains; the prior scratch-file leak is resolved because cleanup now encloses both candidate import and database construction.

| Filename | Overview |
|----------|----------|
| src/lib/persistence/chat-db-worker.ts | Adds the sqlite-wasm survey operation and extends scratch-file cleanup across import and construction failures. |
| src/lib/persistence/owner-host.ts | Replaces sql.js source surveying with the worker-owned survey operation before migration. |
| src/lib/persistence/legacy-blob-reader.ts | Removes the sql.js engine and retains only the legacy IndexedDB blob read. |
| src/lib/persistence/protocol.ts | Adds the typed survey request and result contracts. |
| src/lib/persistence/__tests__/migration-verification.test.ts | Updates migration tests to provide survey results through the stub worker. |
| src/lib/__tests__/architecture-boundaries.test.ts | Enforces that runtime sql.js imports remain confined to the legacy fallback. |

</details>

```mermaid
sequenceDiagram
  participant Host as Persistence owner
  participant Reader as Legacy blob reader
  participant Worker as SQLite worker
  participant Scratch as Scratch database
  Host->>Reader: Read legacy IndexedDB bytes
  Reader-->>Host: Uint8Array
  Host->>Worker: surveyDb(bytes)
  Worker->>Scratch: Import candidate
  Worker->>Scratch: Open and read counts/version/integrity
  Worker->>Scratch: Close and unlink
  Worker-->>Host: SurveyResult
  Host->>Worker: importDb(bytes)
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(persistence): unlink the scratch dat..."](https://github.com/shishir435/ollama-client/commit/763a8bc482b33b705907ed44a06052a4a0504e9d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48785678)</sub>

**Context used:**

### DIFF
--- a/src/lib/__tests__/architecture-boundaries.test.ts
+++ b/src/lib/__tests__/architecture-boundaries.test.ts
@@ -72,6 +72,38 @@ describe("architecture import boundaries", () => {
     expect(rootStores).toEqual([])
   })
 
+  /**
+   * sql.js exists for one job: serving history to a profile still on the legacy
+   * blob backend. Everything else — surveying a source database, verifying an
+   * import, restoring a backup — runs on official sqlite-wasm, which reads the
+   * same file format.
+   *
+   * A new runtime import here is how the second engine comes back, and it comes
+   * back permanently: once two engines read the same file, verification is
+   * comparing engines instead of checking a migration.
+   *
+   * The three dev-only entrypoints are exempt because `config/wxt-hooks.ts`
+   * strips them from store builds. The migration-verification harness is the one
+   * place sql.js is still the right tool: its fixtures have to *write* blobs in
+   * the old topology, which is all sql.js was ever needed for. Type-only imports
+   * in the migration runner do not pull the runtime in.
+   */
+  it("keeps the sql.js runtime to the legacy fallback alone", () => {
+    const allowed = new Set([
+      "lib/sqlite/legacy-db.ts",
+      "entrypoints/benchmark/main.ts",
+      "entrypoints/spike-opfs/main.ts",
+      "entrypoints/persistence-verify/main.ts"
+    ])
+    const offenders = productionSources.filter((file) => {
+      if (allowed.has(file)) return false
+      const source = readFileSync(join(sourceRoot, file), "utf8")
+      return importsModule(source, /sql\.js\/dist\/sql-wasm\.js/)
+    })
+
+    expect(offenders).toEqual([])
+  })
+
   it("keeps application and infrastructure layers independent of features", () => {
     const lowerLayerRoots = ["application/", "background/", "lib/"]
     const offenders = productionSources.filter((file) => {
--- a/src/lib/persistence/__tests__/migration-verification.test.ts
+++ b/src/lib/persistence/__tests__/migration-verification.test.ts
@@ -1,17 +1,18 @@
 import { beforeEach, describe, expect, it, vi } from "vitest"
-import type { ImportResult } from "../protocol"
+import type { ImportResult, SurveyResult } from "../protocol"
 
 // Migration verification and its receipt. Each test re-imports owner-host,
 // because `ensureMigrated` memoizes the attempt for the life of the module —
 // the same reason a real owner runs it once per host.
 
 const legacyBlob = vi.hoisted(() => ({
-  readLegacyBlobBytes: vi.fn(),
-  countLegacyRows: vi.fn()
+  readLegacyBlobBytes: vi.fn()
 }))
 vi.mock("../legacy-blob-reader", () => legacyBlob)
 
 const importResults: ImportResult[] = []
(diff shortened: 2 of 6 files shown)


Shishir435:
Fixed in 763a8bc4 — the finding was right, and the same gap existed in the pre-existing path next to it.

`new pool.OpfsSAHPoolDb(PROBE_PATH)` sat outside the `try`, so a blob that imported but would not open left the scratch file linked. `pool.importDb` throwing partway had the same effect. The cleanup now wraps both.

**The same bug was in `verifyImportCandidate`, and it mattered more there.** Its caller's `finally { pool.unlink(PROBE_PATH) }` only began *after* `db.close()` and `stageRollbackCopy`, so a backup restore whose payload failed to import or open leaked the probe file too — on the path users reach through Restore, not just migration. Cleanup now lives with the function that creates the file, which made the two now-redundant unlinks in the `importDb` case removable and the ownership obvious.

Impact was bounded either way: both paths `unlink(PROBE_PATH)` on entry, so a leak lasted until the next import or survey. But the pool has fixed capacity, so a leaked file is a slot the next import has to grow the pool to replace — `ensureFreeSlots` would paper over it by expanding, which is the wrong direction.

No unit test: both paths need a real sqlite-wasm pool, and the worker is stubbed in unit tests by construction. The packaged persistence gates cover the success paths; "imports but refuses to open" is not a fixture that exists yet, and inventing one is a bigger change than this fix.

Gates: `typecheck`, `lint:check`, `format:check` ✅ · `test:run` ✅ 304 files, 2506 tests

---

One thing to flag while you have this open: #231 is stacked on this branch. It carries the 0.12.6 fixes from #229 onto `release/0.13.x`, which does not have them — without it, 0.13.0 ships the `chunk_feedback` gap and the row-count disclosure that 0.12.6 just fixed. Merge this one first.

PR #231, opened 2026-07-30
Carry the 0.12.6 migration fixes onto 0.13.x

Shishir435:
**Stacked on #230** — branched from `persistence/sqlite-wasm-legacy-reader` so the worker changes don't collide. Merge #230 first; this then reduces to the items below.

## Why this exists

#229 landed fixes on the 0.12.6 line that `release/0.13.x` does not have. Checked each against the branch:

| Fix | On `release/0.13.x` before this PR |
|---|---|
| Receipt observability (`summarizeMigrationReceipt`) | missing |
| Schema applied on every open | missing |
| Destination-completeness guard (`findMissingDurableTables`) | missing |
| Shortfalls instead of row counts (`short by`) | missing |

Without this, **0.13.0 would ship the defects 0.12.6 fixed**: `chunk_feedback` still absent on profiles created before 0.10.0, and row counts still leaving the device in support reports. A forward-port gap, not new work — worth stating plainly, because nothing would have surfaced it until a user on 0.13.0 filed a report with their message counts in it.

## The fixes

1. **Receipt observability.** The receipt was device-local with no reader outside `src/lib/persistence`, so a profile that failed verification looked identical in a report to one that never migrated. The diagnostics bundle and issue draft now carry outcome, attempts, source schema version, integrity verdicts, orphan rows, per-table shortfalls, and failure text.
2. **Schema applied on every open.** `SCHEMA_SQL` ran only when `sessions` was absent, so a table added without a paired migration never reached a profile that already had chats — `chunk_feedback` arrived that way in 0.10.0. All 16 statements are `IF NOT EXISTS`, so it now runs unconditionally; the version stamp stays gated on a fresh database. Plus the destination-completeness guard, so the next such table fails the migration instead of shipping a profile that raises "no such table".
3. **Shortfalls instead of row counts.** Both the diagnostics summary and the verification error formatted absolute count pairs, and both reached a submitted report. Now `table short by N`; absolutes stay in the receipt's structured fields, which never leave the device.

## One extra commit, and why it is here

`pnpm bundle:check` failed on the forward-port: background gzip **193009 > 193000**, over by 9 bytes.

The cause was worth fixing rather than papering over. The background imported `getSafeClientEnvironment` from `error-report`, which pulled that entire module — issue templates, URL length budgeting, the privacy tail, every draft builder — into a chunk with a size budget and no use for any of it. Adding the migration summary to the draft was simply what tipped it.

The helper moves to `src/lib/client-environment.ts`; `error-report` re-exports it so existing importers are unchanged. Background gzip **193009 → 192540** — under budget *with* the new code, instead of raising the ceiling to fit it.

## Deliberate differences from the 0.12.6 commits

- **`DURABLE_TABLES` keeps all ten tables**, not the seven on that line. The guard and both fixtures were extended to `turn_runs`, `ingestion_runs`, `model_pull_runs`.
- **The changelog entry stays with 0.12.6**, the release that ships these to users.

## Gates

`typecheck`, `lint:check`, `format:check` ✅ · `test:run` ✅ 304 files, 2512 tests · `package` + `bundle:check` ✅

<h3>Greptile Summary</h3>

This forward-port restores migration completeness and privacy safeguards from the 0.12.6 line.
- Applies the idempotent schema on every database open while preserving version-stamp semantics for existing profiles.
- Surveys and verifies legacy imports with sqlite-wasm, checks all durable destination tables, and guarantees scratch-file cleanup.
- Adds sanitized migration receipts to diagnostic bundles and issue drafts without exposing absolute history counts.
- Moves client-environment detection into a lightweight module to keep the background bundle within budget.

The PR appears safe to merge.

The previously reported legacy receipt-count disclosure is no longer present: absolute count fields are omitted, count-bearing legacy text is sanitized, and the diagnostics issue draft consumes only the reduced migration summary.

| Filename | Overview |
|----------|----------|
| src/lib/diagnostics/diagnostics-service.ts | Adds a bounded migration-receipt summary that excludes structured counts and sanitizes legacy count-bearing text. |
| src/lib/error-report.ts | Includes sanitized migration evidence in diagnostic issue drafts and preserves the existing environment-helper export. |
| src/lib/persistence/chat-db-worker.ts | Applies the idempotent schema on every open, adds read-only source surveying, and centralizes scratch-file cleanup. |
| src/lib/persistence/owner-host.ts | Uses worker-based source surveys and rejects imports missing any required durable destination table. |
| src/lib/persistence/durable-tables.ts | Adds destination-completeness checks and privacy-preserving mismatch descriptions. |
| src/lib/persistence/protocol.ts | Extends the worker protocol with the typed read-only database survey operation. |
| src/protocol/diagnostics-rpc.ts | Extends the diagnostics contract with sanitized migration evidence. |

</details>

```mermaid
flowchart LR
  Legacy[Legacy SQLite blob] --> Survey[Survey with sqlite-wasm]
  Survey --> Import[Import into OPFS scratch database]
  Import --> Verify[Verify integrity, row shortfalls, and durable tables]
  Verify -->|Pass| Adopt[Adopt OPFS database]
  Verify -->|Fail| Receipt[Persist failure receipt]
  Adopt --> Receipt
  Receipt --> Sanitize[Sanitize diagnostic summary]
  Sanitize --> Bundle[Support bundle and issue draft]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(diagnostics): treat persisted receip..."](https://github.com/shishir435/ollama-client/commit/cb2826b6bb230ac7bc8abfb033ad6f64203505cc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48794687)</sub>

### DIFF
--- a/src/lib/__tests__/architecture-boundaries.test.ts
+++ b/src/lib/__tests__/architecture-boundaries.test.ts
@@ -72,6 +72,38 @@ describe("architecture import boundaries", () => {
     expect(rootStores).toEqual([])
   })
 
+  /**
+   * sql.js exists for one job: serving history to a profile still on the legacy
+   * blob backend. Everything else — surveying a source database, verifying an
+   * import, restoring a backup — runs on official sqlite-wasm, which reads the
+   * same file format.
+   *
+   * A new runtime import here is how the second engine comes back, and it comes
+   * back permanently: once two engines read the same file, verification is
+   * comparing engines instead of checking a migration.
+   *
+   * The three dev-only entrypoints are exempt because `config/wxt-hooks.ts`
+   * strips them from store builds. The migration-verification harness is the one
+   * place sql.js is still the right tool: its fixtures have to *write* blobs in
+   * the old topology, which is all sql.js was ever needed for. Type-only imports
+   * in the migration runner do not pull the runtime in.
+   */
+  it("keeps the sql.js runtime to the legacy fallback alone", () => {
+    const allowed = new Set([
+      "lib/sqlite/legacy-db.ts",
+      "entrypoints/benchmark/main.ts",
+      "entrypoints/spike-opfs/main.ts",
+      "entrypoints/persistence-verify/main.ts"
+    ])
+    const offenders = productionSources.filter((file) => {
+      if (allowed.has(file)) return false
+      const source = readFileSync(join(sourceRoot, file), "utf8")
+      return importsModule(source, /sql\.js\/dist\/sql-wasm\.js/)
+    })
+
+    expect(offenders).toEqual([])
+  })
+
   it("keeps application and infrastructure layers independent of features", () => {
     const lowerLayerRoots = ["application/", "background/", "lib/"]
     const offenders = productionSources.filter((file) => {
--- a/src/lib/__tests__/error-report.test.ts
+++ b/src/lib/__tests__/error-report.test.ts
@@ -191,6 +191,69 @@ describe("buildDiagnosticIssueUrl", () => {
     expect(body).toContain("**Privacy**")
   })
 
+  it("says nothing about migration when the profile never had a legacy blob", () => {
+    const body = bodyOf(buildDiagnosticIssueUrl(bundle))
+
+    expect(body).not.toContain("Chat migration")
+  })
+
+  it("reports a failed migration in the detail a maintainer needs", () => {
+    const body = bodyOf(
+      buildDiagnosticIssueUrl({
+        ...bundle,
+        storage: {
+          ...bundle.storage,
+          backend: "legacy",
+          migration: {
+            outcome: "verification-failed",
+            attempts: 3,
+            recordedAt: 1,
+            extensionVersion: "0.12.6",
+            sourceSchemaVersion: 11,
+            importedIntegrity: "ok",
+            foreignKeyViolations: 2,
+            mismatches: ["messages short by 5"],
+            failure: "Migration verification failed: messages short by 5"
+          }
+        }
+      })
+    )
+
+    expect(body).toContain("Chat migration: verification-failed (attempt 3")
+    expect(body).toContain("schema v11")
+    expect(body).toContain("messages short by 5")
+    expect(body).toContain("Migration orphan rows: 2")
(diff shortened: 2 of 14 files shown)


Shishir435:
Fixed in cb2826b6. The finding was right, and checking it turned up a second instance of the same class that I had introduced.

## 1. Legacy receipt text — confirmed

A receipt is written once per migration and then persists, so `summarizeMigrationReceipt` reads text produced by whichever version performed the attempt. Builds from `release/0.13.x` formatted verification failures as count pairs, and that string went to the bundle and the issue draft verbatim.

The population is narrow — 0.13.x has not shipped, so this is dev and CI profiles — but the objection stands on principle: **the summary is the boundary, and it was trusting its own input.** Structured fields were already re-derived; free-form text was not.

Now stripped:

| Pattern | Why |
|---|---|
| `39199/39204` | the old formatter's shape |
| any standalone number ≥ 4 digits | `integrity_check` prints rowids freely (`row 39204 missing from index`), a rowid is a lower bound on row count, and that output can be embedded in a failure message |

Four digits is the cut so the numbers worth reading survive: shortfalls, attempt counts, schema versions. A shortfall in the thousands gets redacted too — a small precision loss on the rarest failures, and the right trade.

That second pattern was needed because my first attempt only handled pairs, and the test I wrote for it failed on exactly the rowid case I had described in the comment but not defended against.

## 2. The same leak in the new format — mine, not flagged

`findTableCountMismatches` reports `imported: 0` when a table arrives with nothing, so `delta === source` and `messages short by 39204` **was the absolute count**. The wording I introduced to avoid disclosing counts disclosed them in exactly the worst case: total data loss.

That case now reads `messages arrived empty` — says more than a number would, and discloses nothing. `describeMismatch` is shared between the receipt's own failure text and the summary, so the two cannot drift apart again.

## Gates

`typecheck`, `lint:check`, `format:check` ✅ · `test:run` ✅ 304 files, 2514 tests · `package` + `bundle:check` ✅

Background gzip **192633**, still under the 193000 ceiling with the added code.

Two new cases assert the properties directly: a receipt carrying old-format text emerges with no `39199` or `39204` anywhere in the serialized summary, and a table that arrived empty is named rather than counted.

PR #232, opened 2026-07-30
Correct the RAG path and the sql.js removal target

Shishir435:
Two documented statements that are now false. Small, but both are the kind a future reader would act on.

**`AGENTS.md` pointed the RAG pipeline at `src/features/chat/rag/`.** It lives at `src/application/context/rag/` — it moved when context building went to the background, because a feature directory cannot own work the background performs. Noticed while working on #228, when the documented path did not exist.

**`compatibility-ledger.json` targeted `0.13.x` for retiring the live sql.js fallback.** That target predates the evidence window having a start date. The window opened when 0.12.6 shipped today and runs 2–3 months, so it cannot close inside 0.13.x. Both sql.js entries now target `0.14.x`.

The removal gate text also now names what actually shipped — destination-completeness verification, and the diagnostics summary that makes receipts readable — plus the window's start date, so nobody has to reconstruct when the clock started.

No code changes.

<h3>Greptile Summary</h3>

Corrects repository guidance and compatibility-planning metadata.
- Updates `AGENTS.md` to consistently locate and assign ownership of the RAG pipeline under `src/application/context/rag/`.
- Moves both sql.js retirement targets to `0.14.x` and clarifies the evidence-window and removal-gate criteria.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| AGENTS.md | Corrects the RAG pipeline path and resolves the previously reported contradictory ownership guidance. |
| compatibility-ledger.json | Defers both related sql.js removal targets and expands the documented removal gate and evidence window. |

</details>

<sub>Reviews (2): Last reviewed commit: ["docs: name the state owner for every dur..."](https://github.com/shishir435/ollama-client/commit/001cc514c67184fd376e4a60af1259cedc08aadf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48809892)</sub>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -180,7 +180,7 @@ Each feature owns its UI, hooks, and — if needed — its Zustand store.
 
 ### RAG / embeddings
 
-- Pipeline: `src/features/chat/rag/` (`rag-pipeline.ts`, `rag-retriever.ts`, `rag-prompt-builder.ts`, `query-classifier.ts`).
+- Pipeline: `src/application/context/rag/` (`rag-pipeline.ts`, `rag-retriever.ts`, `rag-prompt-builder.ts`, `query-classifier.ts`), driven by `src/application/context/build-context.ts`. It moved out of `src/features/chat/` when context building went to the background — a feature directory cannot own work the background performs.
 - **All** file, memory, and live-page splitting goes through `src/lib/embeddings/chunker.ts`. Do not build a parallel text splitter.
 - Plumbing: `src/lib/embeddings/` (`embedding-strategy.ts`, `embedder-factory.ts`, `hnsw-index.ts`, `keyword-index.ts`, `storage.ts`, `chunker.ts`, `search.ts`).
 - Embedding strategy chain: provider-native → shared model → Ollama fallback.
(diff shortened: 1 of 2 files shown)


Shishir435:
Both fixed in 001cc514.

## The contradiction

The feature table credited `features/chat/` with a `rag/` directory while the section three below it correctly pointed at `src/application/context/rag/`. `features/chat/` actually contains `components hooks lib stores utils` — no `rag/`. The row now says so explicitly rather than just dropping the mention, so the next reader looking for retrieval is sent to the right place instead of finding an absence.

## Plus the ownership item this PR was adjacent to

While here I closed the last open item in the from-scratch audit's rolling work — "document which state system owns each durable or ephemeral value".

Worth recording what the investigation actually found, because I had it wrong when I first scoped it:

- I said ~30 storage keys were missing descriptors. **Wrong.** There are 62 `STORAGE_KEYS` and 62 descriptors — coverage is complete. The 25 I counted were `MESSAGE_KEYS`/`PROVIDER_MESSAGE_KEYS`, which name runtime ports and hold no state.
- I proposed adding a contract test for descriptor coverage. **Already exists** — `storage-key-registry.test.ts` asserts the registry and `STORAGE_KEYS` match exactly.

So the real gap was neither of those. It was the **cross-system** view: the registry answers "does this sync", covers only `chrome.storage`, and says nothing about the other three systems. Four systems hold live state — SQLite, Dexie, `chrome.storage`, Zustand — and nothing said which owns what. That is how a value ends up written from two places with no rule for which wins.

Added: a row per system with what it owns and what it must never hold, plus the rules that follow. Two are already enforced by tests. Also recorded that `stores/theme.ts` and `stores/shortcut-store.ts` are the only durable-backed Zustand stores — the other eleven are ephemeral, and adding a durable value to one is exactly the mistake the section exists to prevent.

Docs only, no code. `lint:check`, `format:check`, `typecheck` ✅ · registry contract test ✅

PR #236, opened 2026-08-07
Make custom providers work without a model list

### LINKED ISSUE #235: Feedback after uninstall
## What were you trying to do?
I was trying to pick a different model from another provider

## What happened?
Cannot see or pick other models

## What could have made you keep Ollama Client installed?
All models from all configured providers must be configured and visible in the dropdown to pick from. 

---
Version: 0.12.7
Language: en-US
Browser: Chrome/Chromium 141 (best effort; edit if incorrect)
OS: macOS (coarse family only)

Shishir435:
Fixes #235.

A custom provider whose `/models` request failed contributed nothing to the model menu, so a provider that was configured, enabled, and working could not be selected from. Hosted gateways that implement `/chat/completions` and nothing else — the reporter's case — hit this on every refresh.

## What was wrong

| | |
|---|---|
| Manual model IDs were dropped when discovery failed | `listModels` merged `customModels` only on the success path, so the ids the user typed in went out with the failed request |
| The connection test *was* the model-list call | An endpoint without a catalog could never pass, so it showed "Connection Failed" while chatting against it worked fine |
| An upstream HTTP 404 was reported as "The provider configuration was not found" | Any `AppError` with `status: 404` got the generic `errors.rpc.not_found` key, which overwrote the provider's own message and described a different problem |
| Providers that returned nothing were silently absent from the menu | `failures` came back over RPC and nothing read it |
| The catalog was re-requested forever | Model-menu refreshes and a 10s health poll each asked again, so a chat-only endpoint collected a 404 every few seconds for an answer that was never going to change |

## What changed

**Declared ids are configuration, not discovery.** `mergeProviderModels` now runs on both paths. A provider whose declared ids carried the list is not reported as a failure; one that produced nothing is, and says which.

**A missing catalog is not an unreachable server.** 404/405/501 from the catalog endpoint means "this server publishes no model list" and is reported through a new `modelListSupported` flag. No answer, 401, 429, and 5xx still fail with their own messages.

**The answer is remembered.** `model-catalog-support.ts` stores it device-local, fingerprinted by wire + base URL + service profile, expiring after a day. A catalog-less provider is asked once; pressing **Test** always re-probes. All discovery goes through one `discoverModels` helper so `listModels` and `testConnection` cannot drift on what counts as a failure.

**The UI says what is true.** Reached-with-no-catalog and declared ids present is a success naming the manual ids; with none it asks the user to add some. A provider running on declared ids gets its own header state instead of a red dot. The model menu names providers that returned nothing.

**The health check costs less.** Once a minute rather than every ten seconds, paused while the page is hidden. Editing a provider still re-checks immediately.

## Also here

- `chore(deps)`: the pre-push audit gate was failing on newly published advisories in the docs toolchain, unrelated to this work. js-yaml and nanoid overrides raised; the two `image-size` advisories have no patched release at any version and arrive through `less` as an optional peer of vite, so they are recorded in `auditConfig` with a note to drop them when a fix exists.
- Version bumped to 0.12.8, CHANGELOG written, provider-setup guide gained an "Endpoints that do not list models" section, and `AGENTS.md` records the discovery rule.

## Verification

`pnpm typecheck`, `pnpm lint:check`, `pnpm test:run` (2462 passing), `pnpm docs:build`, and the full pre-push gate. 18 new/updated tests cover each fix, including that a second discovery pass makes no second request and that pressing Test re-probes anyway.

Checked by hand against `https://api.trustedrouter.com/v1`, which returns 404 for `/v1/models`: the provider tests green, the declared model reaches the menu, chat streams, and the 404 loop is gone.

<h3>Greptile Summary</h3>

This PR allows custom providers without model catalogs to remain usable through manually declared model IDs while distinguishing catalog support from confirmed chat reachability.

- Merges declared model IDs into provider results even when discovery fails.
- Caches catalog support by provider configuration and explicitly probes chat during user-triggered connection tests.
- Updates provider health polling, model-menu diagnostics, RPC errors, documentation, translations, tests, and dependency overrides.

The PR appears safe to merge.

No blocking failure remains; the previously reported false-reachability path now requires an explicit chat probe before a catalog-less draft configuration is reported as usable.

| Filename | Overview |
|----------|----------|
| src/lib/providers/provider-rpc-service.ts | Centralizes discovery, preserves declared models, and verifies catalog-less draft configurations through the chat endpoint before claiming reachability. |
| src/lib/providers/model-catalog-support.ts | Adds device-local, fingerprinted, expiring catalog-support state with explicit invalidation behavior. |
| src/features/model/hooks/use-provider-settings-state.ts | Presents distinct connection-test messaging for catalog-less providers and manually declared model IDs. |
| src/features/model/hooks/use-provider-health.ts | Reduces background polling frequency, pauses hidden-page checks, and avoids representing unverified catalog-less providers as connected. |
| src/features/model/hooks/use-provider-models.ts | Consumes structured model-list failures and exposes providers that contributed no usable models. |
| src/protocol/provider-rpc.ts | Extends the validated RPC result contracts with catalog-support and provider-failure metadata. |
| src/background/rpc-server.ts | Preserves provider-specific user messages instead of replacing upstream 404 diagnostics with a configuration-not-found translation. |

</details>

```mermaid
flowchart TD
  A[Discover models] --> B{Catalog response}
  B -->|Models returned| C[Merge declared model IDs]
  B -->|404 / 405 / 501| D[Remember catalog unsupported]
  B -->|Other failure| E[Report provider failure]
  D --> C
  C --> F{Explicit Test?}
  F -->|No| G[Return models and catalog state]
  F -->|Yes, model available| H[Probe chat endpoint with one token]
  F -->|Yes, no model available| I[Ask user to declare a model ID]
  H -->|Success| J[Report usable provider]
  H -->|Route absent| K[Clear cached state and report bad base URL]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(providers): confirm a catalog-less e..."](https://github.com/shishir435/ollama-client/commit/1a8dd9d2b6b3fc3f3c83d49513e4dd66e9e8e9e4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51331447)</sub>

**Context used:**


Shishir435:
Good catch — fixed in 1a8dd9d2.

You're right that a status-only check cannot tell a chat-only gateway from a mistyped base URL: both answer 404 on `/models`. Reporting the second as reachable was wrong, and caching it made it worse.

**An explicit test now confirms against the endpoint the user actually needs.** When the catalog is absent and at least one model ID is declared, `testConnection` streams a single token from `/chat/completions` with that ID and aborts at the first byte:

- it answers → reachable, `modelListSupported: false`, models come from the declared IDs;
- the chat route is missing too (404/405/501) → thrown as a base-URL problem ("Nothing answered at {url}: neither a model list nor a chat endpoint is there. Check the base URL — hosted providers usually need the version suffix, such as /v1"), and the recorded catalog answer is **cleared**, so a corrected URL gets a clean probe rather than a day of silence;
- 401/429/5xx → rethrown as itself, upstream diagnostic intact. That was the second half of your comment: the original code swallowed the error, and it no longer does.

**The background check does not do this.** A `stored` (health-poll) test never sends a chat request and now reports `reachable: false` rather than claiming something it did not verify — it is a health poll, not a licence to spend inference on a metered endpoint. With no declared IDs there is nothing to probe with, so the test reports "no model list, add model IDs" instead of a success, which is what the UI already showed for that case.

The 24h cache still exists — its job is to stop a *pointless repeated request*, and being wrong there costs nothing but a missing catalog — but it can no longer be the basis for a reachability claim.

Three tests added: the probe runs with the declared ID and `max_tokens: 1`; a missing chat route surfaces as a base-URL error and clears the marker; the background path spends no chat request and claims no reachability.

PR #237, opened 2026-08-07
Stop testing the connection on every keystroke

Shishir435:
Two more for 0.12.8, both in the same family as #236: requests nobody asked for.

## Typing a base URL ran a connection test per keystroke

`updateConfig` maps over provider state on every keypress, producing a new `providers` array; `useProviderHealth`'s effect keyed on that array's identity. Measured against the real settings hook before the fix:

```
TYPED 32 chars -> 32 connection tests
```

On Ollama that is 32 `/api/tags`. On a hosted provider it is 32 billable calls to edit one field — and each one runs against a half-typed URL.

The effect now keys on what actually decides the answer: which providers exist, which are enabled, and where they point. Renaming a provider, editing its declared model IDs, or retyping an API key cannot change whether it is reachable, so none of them cost a request. A change that *does* matter settles for 700ms first, so a burst of keystrokes collapses into one check; the screen's first check still runs immediately.

Two regression tests: a re-render that changes only the name spends nothing, and typing a full URL character by character produces exactly one check after the burst.

## Support reports called a working provider unreachable

`providerReachable` was `failures.length === 0`, and the `model_list_unsupported` code introduced in #236 counts as a failure. So a chat-only provider that works filed a report saying "Provider reachable: no", pointing whoever read it at the wrong thing. That code is now excluded — it means "nothing to discover", not "could not reach".

This one is a regression from my own work earlier in 0.12.8, so it belongs in this release rather than the next.

## Also

The three discovery failure codes are named once in `provider-rpc.ts` instead of being spelled out on both sides of the RPC boundary, where a typo would have been a silent behaviour change rather than a build error.

## Verification

`pnpm typecheck`, `pnpm lint:check`, `pnpm test:run` (2466 passing), and the full pre-push gate.

<h3>Greptile Summary</h3>

The PR changes provider-health polling to react to successful stored-configuration revisions rather than draft edits, and centralizes model-discovery failure codes.

- Prevents draft URL keystrokes from repeatedly testing the stored endpoint.
- Refreshes provider health after successful saves and avoids concurrent polling sweeps.
- Corrects support-report reachability for providers without model-list support.
- Reuses typed discovery-failure constants across the RPC boundary.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/features/model/hooks/use-provider-health.ts | Replaces provider-array identity effects with a saved-state signature and serializes overlapping health sweeps. |
| src/features/model/hooks/use-provider-settings-state.ts | Tracks successful provider persistence with savedRevision so stored endpoint health is refreshed after saves. |
| src/features/chat/components/chat-error-report-action.tsx | Treats model-list-unsupported responses as reachable when producing support diagnostics. |
| src/protocol/provider-rpc.ts | Defines shared constants and a type for model-discovery failure codes. |
| src/lib/providers/provider-rpc-service.ts | Uses shared discovery-failure constants without changing the RPC result shape. |

</details>

```mermaid
sequenceDiagram
  participant UI as Provider Settings
  participant State as Settings State
  participant RPC as Provider RPC
  participant Health as Health Hook
  UI->>State: Edit draft base URL
  Note over State,Health: Draft edits do not change savedRevision
  UI->>State: Save configuration
  State->>RPC: Persist provider config
  RPC-->>State: Saved provider
  State->>State: Increment savedRevision
  State->>Health: New saved signature
  Health->>RPC: Test stored provider
  RPC-->>Health: Current endpoint health
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(providers): check health when a save..."](https://github.com/shishir435/ollama-client/commit/d6ad907ed3e31f1ad1fc5c8d8966654030210294) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51339719)</sub>

**Context used:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,10 +48,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - A provider running on model IDs you declared is labelled as such rather than
   reported as a failed connection.
 - Editing a provider's base URL no longer runs a connection test on every
-  keystroke. Typing one URL fired one test per character, each against the
-  half-typed value — on a hosted provider, a real API call per keypress. Edits
-  now settle first, and a re-render that changes nothing about where a provider
-  points costs nothing at all.
+  keystroke. Typing one URL fired one test per character — on a hosted
+  provider, a real API call per keypress — and none of them told you anything,
+  because the check reads saved configuration and you were still typing. The
+  check now runs when a change is saved, and the status you see after saving
+  describes the endpoint you just saved rather than the one before it.
 - A support report from a provider that publishes no model list no longer says
   "Provider reachable: no". It is reachable; it just has nothing to discover,
   and the old wording sent whoever read the report after the wrong thing.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   provider still re-checks it immediately.
 - A provider running on model IDs you declared is labelled as such rather than
   reported as a failed connection.
+- Editing a provider's base URL no longer runs a connection test on every
+  keystroke. Typing one URL fired one test per character, each against the
+  half-typed value — on a hosted provider, a real API call per keypress. Edits
+  now settle first, and a re-render that changes nothing about where a provider
+  points costs nothing at all.
+- A support report from a provider that publishes no model list no longer says
+  "Provider reachable: no". It is reachable; it just has nothing to discover,
+  and the old wording sent whoever read the report after the wrong thing.
 
 ## [0.12.7]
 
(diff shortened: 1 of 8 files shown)


Shishir435:
Right on both counts — fixed in d6ad907e, and the fix made the hook simpler rather than more complex.

The debounce was papering over the wrong key. The check tests **stored** config, so keying it on a draft base URL was wrong twice over: it fired against the old stored value while the user typed, and then — exactly as you describe — the save landed without changing the signature, so the endpoint that had replaced it was never checked until the next heartbeat.

**The key is now what the check actually reads.** The effect depends on which providers are enabled, plus a `savedRevision` the settings hook bumps when a save, add, remove, or enable-toggle completes. So:

- typing costs nothing, because a draft URL describes an endpoint this check would not contact either way;
- a save re-checks immediately, against what it just saved — the stale-status window you found is closed;
- the 700ms settle window is gone entirely. It only existed to hide the wrong key.

**The duplicate request is gone too.** A sweep can no longer overlap itself, so the visibility handler and the interval landing together run one pass across the providers instead of two concurrent ones.

Measured end to end against the real settings hook: a 32-character base URL is still 0 connection tests, and a save is 1. Test coverage follows the same split — one test that typing spends nothing, one that a bumped revision re-checks straight away.

Worth noting the first version of this would have passed review on the numbers alone (0 requests while typing is what I set out to prove) while quietly making the post-save status worse. Thanks for reading past the metric.

PR #257, opened 2026-08-10
perf(models): stop doing idle work while a panel is open

Shishir435:
Two things ran on their own whether or not anyone asked, and both showed up as sustained extension-process CPU with an idle sidepanel.

## Diagnostics collected on mount

`ChatErrorReportAction` prefetched the support bundle in a mount effect. That bundle runs the whole diagnostic suite behind it — a storage write probe, a SQLite transaction, message and vector counts, and a provider catalog request — and the component mounts for every failed message, on every sidepanel open, and on every pass of the virtualized message list. Closing and reopening the panel re-ran all of it.

It is now collected when the pointer or focus reaches one of the two buttons that need it. A click still finds it ready, which matters because `clipboard.writeText` depends on the click's transient activation and a collection started on click could outlive it — and a session that never presses either button never pays for it at all.

## Provider catalogs refetched every 10 seconds

`ProviderStatusIndicator` called `refresh()` on a 10-second interval. `refresh()` invalidates and refetches, so it bypassed the query cache and re-ran discovery against every enabled provider — hidden panel included. For a hosted router with hundreds of models that is a real request against someone else's rate limit six times a minute, plus the re-parse and re-render behind it.

The catalog queries now carry the interval themselves. That means one refetch per provider on a user-set schedule, paused while the surface is hidden (react-query does not poll in the background), and skipped entirely while the cached list is still fresh.

## The new setting

**Providers → Model list refresh**: off, 30 seconds, 1 minute, 2 minutes, 5 minutes; default 1 minute. Sync-safe, so it follows the user — and floored at 30 seconds on read, because a value can arrive from another device running an older build. Pulling, deleting or editing a provider still refreshes immediately, and the refresh button still forces one.

## Tests

- `src/features/model/lib/__tests__/catalog-refresh.test.ts` — normalization, the floor, the off state, and the stale-time relationship.
- `src/features/model/components/__tests__/provider-status-indicator.test.tsx` — no timer of its own, still refreshes on demand.
- `chat-error-report-action.test.tsx` updated: mounting a failed message must not call `DiagnosticsGetBundle`; pressing the button must.

## Gates

`pnpm typecheck`, `pnpm lint:check`, `pnpm test:run` (332 files / 2756 tests), `pnpm check:dead`, `pnpm check:i18n`, `pnpm generate:resources`, `pnpm build` all pass. New copy is translated in all nine locales.

<h3>Greptile Summary</h3>

This PR removes unconditional diagnostics collection and provider-catalog polling, replacing them with demand-driven diagnostics and configurable query-level refresh behavior.
- Defers support-bundle collection until a diagnostics action receives pointer, focus, or click intent.
- Uses a promise-backed clipboard item where supported to preserve the click’s transient activation.
- Moves catalog polling into per-provider queries with a normalized, synchronized refresh setting.
- Adds registry, component, polling, normalization, storage, and locale coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/features/chat/components/chat-error-report-action.tsx | Defers expensive diagnostics work and addresses the previously reported clipboard-activation delay on ClipboardItem-capable targets. |
| src/features/model/components/provider-settings.tsx | Adds a catalog-refresh selector whose displayed value uses the same normalization as polling. |
| src/features/model/hooks/use-provider-models.ts | Replaces forced global refreshes with configurable per-provider React Query intervals and aligned stale times. |
| src/features/model/lib/catalog-refresh.ts | Defines the supported intervals and safely normalizes arbitrary synchronized values to an offered choice. |
| src/features/settings/registry/models.ts | Registers the new setting for search and deep-link resolution with matching mounted-control coverage. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(models): make the refresh interval a..."](https://github.com/shishir435/ollama-client/commit/0cd7c2619070b18c72584e4d0590795fff530405) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51905126)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Settings & Options Page](https://app.greptile.com/shishir435/-/custom-context/knowledge-base/shishir435/ollama-client/-/docs/settings-options.md)
</details>


Shishir435:
All three addressed in 0cd7c261.

**P1 — stored interval breaks selection.** Correct, and the divergence was the worse half: the control bound the raw stored value while `useProviderModels` polled on the normalized one, so a synced 45s rendered a blank trigger over an interval that really was in force. Clamping alone would not have fixed it either — `normalizeCatalogRefreshMs(45_000)` returned 45000, still not an option.

This setting is an enum wearing a number, so normalization now snaps to the nearest offered choice rather than clamping, with ties going to the slower one (fewer requests is the point of the setting). The control binds `normalizeCatalogRefreshMs(stored)` — the same value the queries poll on, so the two cannot disagree by construction. New cases in `catalog-refresh.test.ts` assert every result is a member of `CATALOG_REFRESH_CHOICES_MS`.

**P1 — diagnostics delay clipboard activation.** Also correct. The hover prefetch covers the common mouse path but not a click that lands before it resolves, or focus-then-Enter, and the diagnostic suite includes a provider catalog request that can take seconds — well past what a click's transient activation survives.

Fixed at the primitive rather than the timing: `clipboard.write` accepts a *promise* of the item data, so the write is claimed inside the click and settled whenever the bundle arrives. `writeText` stays as the fallback where `ClipboardItem` does not exist, which is the old behavior and no worse than before. Two tests now — the promise-valued path asserts the copied text by reading it back off the `ClipboardItem`, and a second deletes the global to exercise the fallback.

**P2 — setting registration lacks coverage.** Fair; AGENTS.md asks for it and I skipped it. `settings-registry.test.ts` now requires search to find `provider-catalog-refresh` by "model list refresh", "polling interval" and "network requests", and `provider-settings.test.tsx` asserts the focus id mounts and its trigger shows a real choice instead of an empty label — which is also the regression guard for the first finding.

Gates green locally: typecheck, lint, 332 files / 2761 tests.

PR #255, opened 2026-08-10
fix(persistence): make owner startup deterministic

Shishir435:
Closes H2 of the `0.13.0` architecture-audit sequence (PR 2 of 8).

## The failure

Registering the persistence topology was a fire-and-forget dynamic import, so nothing could wait for it. Startup tasks that touch the chat database ran immediately: on a cold Chromium worker the ensure hook was not installed yet, the client fell back to runtime messaging, and a service worker never receives its own message — so the call sat there until the 30s client timeout with a working owner one tick away.

## What changed

- **One readiness promise, owned by the composition root.** `src/background/persistence-readiness.ts` starts the topology and hands the promise to `initializeBackgroundStartup`.
- **An owner-ready handshake.** `createDocument()` resolving proves a document exists, not that its listener, worker, WASM binary or backend choice are up. `ensurePersistenceOwnerReady()` proves the whole chain with one `ping`, retrying the window where the host page has not evaluated its script yet. The proof is cached per owner instance and re-taken when the document is recreated; a rejection is never cached, so a later caller retries instead of inheriting a failed boot. Concurrent callers share one document and one handshake.
- **Ordered startup.** Lifecycle flags → owner → data-shape recovery (backup import, provider migration, embedding-dimension migration, sequential) → durable workflow recovery, two at a time. Backup import can replace the whole database; resuming a turn against one mid-replacement was previously possible. A boot that cannot reach the owner skips database startup once and says so, instead of six tasks each waiting out their own timeout.
- **Retry policy with evidence.** A request that failed before anything was sent throws `PersistenceNotDeliveredError` and may be repeated whatever it is, because it provably did not execute. A request that failed after hand-off keeps the old rule and is repeated only when the op is idempotent by construction.

`registerChromiumPersistenceControl` no longer starts the owner itself, and `PERSISTENCE_ENSURE` from extension pages now waits for an owner that can serve rather than one that merely exists.

## Tests

14 new across three files:

- `src/lib/persistence/__tests__/owner-readiness.test.ts` — handshake retry, coalescing, cached proof, uncached failure, re-proof after recreation, owner answering with a failure.
- `src/lib/persistence/__tests__/client-delivery.test.ts` — a write the owner never received is repeated; a write of unknown outcome is not; an idempotent op still is.
- `src/background/__tests__/startup-database-order.test.ts` — nothing touches the database before readiness, data-shape recovery completes before workflow recovery, concurrency stays at 2, a failed workflow does not cancel the others, and both skip paths log.

## Gates

[...]

PR #259, opened 2026-08-10
fix(turns): make cancellation and lifecycle transitions durable

Shishir435:
Closes H3 of the `0.13.0` architecture-audit sequence (PR 3 of 8).

## The failure

Stopping a turn wrote nothing durable. `port-router.ts` aborted the in-memory controller and returned; the row still read `generating`. A worker that died before the generation noticed the abort restarted, found a live turn, and handed it straight back to the provider — the user's stop undone by a crash, and a metered endpoint billed twice for a request they cancelled.

## Cancellation is committed intent

A new `cancelling` state lands durably **before** the controller is aborted. Recovery does not resume it, and startup finalizes an interrupted one to `cancelled` without reissuing any provider work. A stop for an id that names no live turn — a selection-action scope key, or a second click — writes nothing.

## Transitions are compare-and-set

`TURN_STATUS_PREDECESSORS` is the whole state machine, and every status write carries the target's allowed predecessors into the `WHERE` clause, so the database decides whether a transition is legal rather than the caller. That holds without an application lock across the two workers that can be mid-flight after a restart:

- a terminal row never regresses
- a duplicate start cannot claim a turn twice
- a generation that finishes after a stop cannot pull a settled row back into flight
- `cancelling` still accepts any terminal, because the abort races the work it aborts and stranding the row would be worse

`updateTurnRun` resolves false on a refusal and `TurnRuntime` treats that as "someone else owns this turn", doing no provider work.

## Failures keep their shape

The terminal stream event already carries a status, kind, message key and incident id. Throwing a plain `Error` built from its text discarded all of it — which is why a provider 500 reached the bubble as a bare "Turn failed before completion." `DurableTurnGenerationError` carries the structured `AppFailure` through the turn row, the assistant row, the reconnect snapshot and the bubble. The generic text is now the last resort for a turn that died before the stream said anything.

## Unreadable rows are quarantined

A turn row that will not parse was skipped by recovery, silently, on every boot forever. It is terminally failed with a diagnostic that names the record and no content.

## `building-context` → `building_context`

The only durable status value that was not one word, and the only transition-table key that had to be quoted. The value is persisted, so this is migration 13 plus both SQL literals and the stream enum. `pnpm verify:opfs-migration` passes with the migration running on top of a seeded v12 database.

## Folded-in hardening

Two items raised while reviewing #255 and #257, kept as their own concerns but shipped here rather than in a separate PR:

- **Startup deadlines.** #255 serialized data-shape recovery, so a step that never settles blocks every later step and all four workflow recoveries. Each task now has a 120s deadline — generous, because a large migration is slow rather than stuck.
- **Clipboard fallback.** Feature-detecting `ClipboardItem` does not prove an engine honors a *promise* as item data. A refusal now falls back to `writeText` instead of reporting failure for a bundle we are holding.

## Tests

- `packages/contracts/src/__tests__/turn-status.test.ts` — terminal rows never move, cancellation intent is neither resumable nor terminal, no backwards transitions, `submitted` unreachable.
- `turn-runs.test.ts` — CAS carries predecessors and refuses a stale transition without flushing, intent commits only for a live turn, a repeated stop is a no-op, interrupted cancellations finalize, recovery never selects `cancelling`, an unreadable row is quarantined with no content in the diagnostic.
- `turn-runtime.test.ts` — a refused claim does no context or provider work; a stop during context building prevents generation.
- `durable-turn-runtime.test.ts` — intent commits on the stop path, a non-turn key answers false, a persistence failure never blocks the abort, finalization runs before recovery reads.
- `chat-error-report-action.test.tsx` — the clipboard falls back when a promised item is refused.

## Gates

`pnpm typecheck`, `pnpm lint:check`, `pnpm test:run` (336 files / 2797 tests), `pnpm check:dead`, `pnpm build`, `pnpm verify:opfs-migration` all pass.

<h3>Greptile Summary</h3>

The PR makes turn cancellation and lifecycle transitions durable while preserving structured generation failures.
- Adds SQL-enforced compare-and-set transitions and durable `cancelling` intent.
- Finalizes interrupted cancellations assistant-first and protects cancelling assistants from stale-message recovery.
- Migrates the persisted `building-context` status to `building_context`.
- Adds startup deadlines and a clipboard fallback for engines that reject promised clipboard items.

The PR appears safe to merge.

No blocking failure remains; the previously reported cancellation-finalization and stale-sweep issues are addressed by assistant-first recovery and durable ownership of assistants while turns remain cancelling, while the startup overlap behavior is explicitly accepted and deferred.

| Filename | Overview |
|----------|----------|
| packages/contracts/src/turns.ts | Defines the authoritative turn state machine, cancellation ownership, resumable states, and legal predecessors. |
| packages/chat-runtime/src/turn-runtime.ts | Treats refused lifecycle claims as lost ownership and prevents duplicate context or provider work. |
| src/lib/repositories/turn-runs.ts | Implements guarded lifecycle writes, durable cancellation intent, interrupted-cancellation recovery queries, and unreadable-row quarantine. |
| src/background/durable-turn-runtime.ts | Preserves structured generation failures and finalizes interrupted cancellations assistant-first before resuming turns. |
| src/lib/repositories/sqlite-chat-history.ts | Excludes assistants owned by cancelling or resumable turns from stale-message recovery. |
| src/lib/sqlite/migrations/rename-building-context-status.ts | Migrates the persisted building-context status spelling to the new contract value. |
| src/background/startup.ts | Adds bounded waits around startup tasks, with non-cancelling abandonment explicitly documented as deferred behavior. |

</details>

```mermaid
sequenceDiagram
  participant UI as Sidepanel
  participant BG as Background
  participant DB as SQLite
  participant P as Provider
  UI->>BG: Stop turn
  BG->>DB: CAS live status to cancelling
  DB-->>BG: committed
  BG->>P: Abort generation
  Note over BG,DB: After worker restart
  BG->>DB: Read cancelling turns
  BG->>DB: Mark assistant done
  BG->>DB: CAS cancelling to cancelled
  BG->>DB: Read resumable turns
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(turns): keep a settling turn&#39;s assis..."](https://github.com/shishir435/ollama-client/commit/4b6841fafe570c5561eafabc011c20dcf74f8de0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51937198)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Chat Flow: Send, Stream, Cancel, Edit, and Fork](https://app.greptile.com/shishir435/-/custom-context/knowledge-base/shishir435/ollama-client/-/docs/chat-flow.md)
</details>


Shishir435:
Both addressed in 9cbbf6b0.

**P1 — cancellation leaves assistant unfinished.** Real, and the worse half of the bug I was fixing: I settled `turn_runs` and left the assistant at `done = 0`, so a response the user explicitly stopped reads as interrupted and the stale-message sweep offers a retry for it. `finalizeInterruptedCancellations` now returns the rows it settled so recovery can finish each assistant, and a failure there does not block the live turns behind it.

**P2 — deadline does not stop the task.** Correct. None of the startup operations takes an `AbortSignal`, so the race abandons rather than cancels. Rather than imply otherwise, the comment and the error now say "abandoned and may still be running", and threading cancellation through backup-import recovery, the two migrations and workflow recovery is tracked as its own item in `RELEASE_ROADMAP.md`. It is a real gap, just a bigger change than this PR.

Gates: typecheck, lint, 336 files / 2800 tests.

Shishir435:
Right, and my previous fix only moved the window. Fixed in 274ca254 — but not with a transaction.

The two writes belong to different repositories (`turn_runs` and the chat-history facade), so they cannot share one. Write **order** carries the guarantee instead: the assistant is finished first and the turn settled second, one turn at a time. A worker that exits in between leaves the row at `cancelling`, which the next boot finds and repeats — and `updateMessage(id, { done: true })` is idempotent, so repeating costs nothing. Settling the turn first is what closed the door with the assistant still unfinished.

Same reasoning for the failure path: a turn whose assistant cannot be finished is left `cancelling` rather than settled over, so it is retried next boot instead of stranded.

This is the pattern the PR already uses for the stop itself — commit the intent that makes the intermediate state recoverable, then act. `finalizeInterruptedCancellations` is now split into a read-only `getInterruptedCancellations` and a per-turn `finalizeCancelledTurn` guarded on `status = 'cancelling'`, and the test asserts the write order rather than just the effects.

Shishir435:
Correct, and it is a hole I opened: I added `cancelling` and never taught the sweep about it. Fixed in 4b6841fa.

Ownership and recovery are now separate lists, and the difference is the point — `RESUMABLE_TURN_STATUSES` excludes `cancelling` because nothing may reissue a turn the user stopped, while the new `TURN_OWNED_ASSISTANT_STATUSES` includes it because its assistant has not been abandoned. `finalizeInterruptedMessages` interpolates the contract's list instead of spelling the statuses out, so the next status added there cannot silently go missing here — which is exactly how this one did.

Test asserts the relationship rather than the literal: ownership contains every resumable status plus `cancelling`, and no terminal one.

PR #262, opened 2026-08-11
refactor(providers): give model discovery one policy owner

Shishir435:
`discoverModels` was private to `provider-rpc-service.ts`, so two production callers bypassed it: tool gating (`resolve-model-tools.ts`) and the embedding-model check both called `provider.getModels()` directly. Neither honoured a remembered catalog absence, so a chat-only gateway got re-asked to 404 on every turn and every embedding check.

The helper moves to `src/lib/providers/model-discovery.ts` and both callers go through it. A boundary test fails on any `.getModels(` outside `lib/providers/`.

Closes H5.

<h3>Greptile Summary</h3>

The PR centralizes provider model discovery policy and routes tool capability resolution and embedding checks through it.
- Extracts remembered catalog-absence and failure handling into `model-discovery.ts`.
- Prevents repeated catalog requests after a settled absence.
- Adds a repository-wide boundary test limiting direct `getModels` calls to the policy owner and wire-adapter delegation.
- Adds focused discovery, embedding, tool-resolution, and test-stability coverage.

The PR appears safe to merge.

No blocking failure remains; both previously reported issues are addressed by the current code.

| Filename | Overview |
|----------|----------|
| src/lib/providers/model-discovery.ts | Extracts the existing discovery policy into the shared owner while preserving absence, failure, force-reprobe, and abort semantics. |
| src/background/handlers/handle-embedding-download.ts | Routes resolved-provider checks through shared discovery and correctly avoids repeating the default Ollama catalog request after a settled absence. |
| src/background/lib/resolve-model-tools.ts | Replaces per-turn direct model listing with shared discovery so remembered catalog absence is honored. |
| src/lib/__tests__/architecture-boundaries.test.ts | Narrows the policy exemption to the discovery owner and legitimate `super.getModels` wire-adapter delegation. |

</details>

```mermaid
flowchart LR
  Callers["RPC listing / tool gating / embedding check"] --> Discovery["model-discovery.ts"]
  Discovery --> Memory{"Remembered catalog absence?"}
  Memory -- Yes --> Absent["Return catalog: absent"]
  Memory -- No --> Provider["provider.getModels()"]
  Provider -- Models --> Present["Record support and return catalog: present"]
  Provider -- 404/405/501 --> RecordAbsent["Record absence and return catalog: absent"]
  Provider -- Other failure --> Failed["Return catalog: failed"]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(providers): stop the embedding check..."](https://github.com/shishir435/ollama-client/commit/adb8a5f7b39feba3613b9ed2d2550065bd0e6966) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52158082)</sub>

**Context used:**


Shishir435:
Both findings fixed in `adb8a5f7`, plus the CI failure.

**P1 — remembered absence re-requests catalog.** Correct, and reachable through one path: an embedding model that is neither the default name nor tied to an explicit provider skips the direct fast path, resolves to Ollama, and lands in the fallback with `resolvedProviderId` equal to the default. Since the fallback only runs for the default provider, the `/api/tags` call it makes is the exact request discovery just made through `provider.getModels` — so a settled verdict was followed by asking the identical endpoint again.

The verdict now travels out of the discovery call and a settled `absent` skips the fallback. An **unset** verdict still runs it: a provider that fails to resolve produces no verdict, and that is the case the fallback was written for. Both are covered by tests, and I mutation-checked the guard.

**P2 — directory-wide exemption.** Agreed, and the failure mode you name is exactly the shape of the two callers this rule was written for. Only `model-discovery.ts` is exempt now, plus `super.getModels` — a subclass delegating to its base wire format rather than a caller skipping the policy. Verified by dropping a `lib/providers/sneaky-service.ts` calling `provider.getModels()` into the tree: flagged by name under the narrow rule, silently passing under the old one.

One thing worth stating rather than leaving implied: the default provider's own `/api/tags` check is still a direct fetch, and stays that way deliberately. It skips provider resolution, and remembered absence exists to stop repeat requests against remote endpoints that charge for them — not against the user's own Ollama on loopback, whose catalog endpoint is not optional. That is now a comment at the call site and a line in AGENTS.md, so it reads as a decision rather than an oversight.

**CI.** The red check was `use-semantic-chat-search`, unrelated to this PR — it slept a fixed 10ms waiting for a React render before asserting synchronously, which a loaded runner can outlast. Confirmed pre-existing (file untouched here, passes locally on repeat). It polls with `waitFor` now.

PR #263, opened 2026-08-11
fix(persistence): decode durable rows and type persistence failures

Shishir435:
Three gaps at the storage boundary, all the same kind — something crossing a boundary was asserted rather than checked.

Durable job repositories claimed row types with `as unknown as Row[]`; they now declare Zod row schemas and go through a shared decoder, with a per-repository failure policy (quarantine / raise / drop-and-log). Persistence failures are typed, and the owner's SQLite text no longer becomes the error message every log line prints. `packages/contracts/src/chat.ts` splits into four concept modules behind an unchanged barrel.

A real-engine smoke test proves each writer still agrees with its own schema.

Closes H7 — the last code PR before H8.

<h3>Greptile Summary</h3>

The PR validates durable SQLite rows at repository boundaries, introduces typed and safely redacted persistence failures, and splits the chat contracts into concept-specific modules behind the existing public barrel.
- Adds shared Zod-based row decoding with repository-specific malformed-row policies.
- Preserves retry semantics while typing owner startup, timeout, response, and execution failures.
- Makes persistence diagnostic detail and causes non-enumerable and private to structured logging.
- Retains the existing public chat contract surface while separating activity, attachment, replay, and message schemas.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/persistence/errors.ts | Defines typed persistence failures with safe messages, retry metadata, and private non-enumerable diagnostics. |
| src/lib/persistence/client.ts | Normalizes both message-based and in-process persistence failures without exposing owner error text or weakening write retry safety. |
| src/lib/persistence/owner-host.ts | Distinguishes startup failures known to precede delivery from worker-stage failures with unknown execution outcomes. |
| src/lib/log-redaction.ts | Adds symbol-based private error keys and redacts declared diagnostic properties and causes. |
| src/lib/repositories/row-decoder.ts | Adds shared runtime decoding and bounded schema-failure logging for durable database rows. |
| src/lib/repositories/turn-runs.ts | Applies decoded turn-row handling and preserves quarantine behavior for malformed durable turns. |
| packages/contracts/src/chat.ts | Replaces the monolithic schema module with an unchanged public barrel over concept-specific contract modules. |

</details>

<sub>Reviews (6): Last reviewed commit: ["fix(persistence): keep owner text out of..."](https://github.com/shishir435/ollama-client/commit/f5968a78e32d4babe8ef45d0443bda1fa791d8fd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52168010)</sub>

**Context used:**


Shishir435:
Fixed in the latest commit — you were right, and about the path that matters most.

The typed errors covered the messaging path and left the in-process fast path alone, which is exactly backwards. `__persistenceHostCall` reaches `callWorker` directly, and the worker posts back `error.message` — SQLite's own text. So that branch rejected with a raw Error carrying the disclosure, no `op`, no `reason`, no `retryable`. On Firefox MV2 the background page is the owner *and* the heaviest client, so the context doing the most persistence work was the only one still getting untyped failures.

Both in-process entry points are wrapped now: `sendOnce`'s host-call branch, and `ensurePersistenceHost` — the latter because other work the owner hosts calls it directly and a raw rejection would escape past `sendOnce` entirely.

One judgment call worth stating: an in-process failure maps to `owner-error`, not `not-delivered`, even though some of these fail before the worker is posted to (`ensureMigrated` rejecting, for instance). The two are indistinguishable from the client, and `not-delivered` is the label that makes a non-idempotent write retryable. Claiming non-execution we cannot prove is what turns a lost write into a duplicated one, so the unprovable case takes the conservative label.

An already-typed error passes through unchanged — re-wrapping would relabel a timeout as an owner error and take its retryability with it.

Three tests on the branch, and I mutation-checked the wrapper: removing the try/catch fails the disclosure test immediately. Gates green, 341 files / 2847 tests.

Shishir435:
Fixed. Verified against `log-redaction.ts:156`, and there was a second door you didn't flag.

**The one you found.** `redactError` iterates `Object.keys(error)` and passes each through `redactProperty`, which only redacts on key-name match — `detail` matches nothing, so it went to the console and the diagnostics bundle verbatim. My design premise was "detail is opt-in for diagnostics"; the generic logger made it opt-out.

**The one next to it.** `redactError` also reads `error.cause` **by name** and recurses. `PersistenceNotDeliveredError` sets the owner's raw Error as its cause, so the same SQLite text leaked a second way — and that path is immune to non-enumerability, since it is not reached by enumeration.

**The fix.** Redaction is keyed on property names, which works when the name predicts sensitivity (`apiKey`, `content`). It cannot work for a value whose sensitivity comes from *where it was obtained* — an error wrapping another process's message knows its text is untrusted, and no key-name list can. So the error declares it, under a `PRIVATE_ERROR_KEYS` symbol: precisely the kind of key `Object.keys` does not report and `JSON.stringify` does not serialize. `redactError` honours it for enumerated properties **and** for `cause`.

`detail` and `cause` are also non-enumerable. The marker covers the redaction path; non-enumerability covers loggers that do not know the convention — `console.log(error)`, a spread, `JSON.stringify`. `error.detail` still reads by name, the only access intended.

Three tests, and I mutation-checked the doors independently: dropping the marker fails the cause-chain test, making `detail` enumerable fails the unknown-logger test. A fourth asserts an ordinary error keeps its cause and own properties — the marker is opt-in, and a redaction change that quietly blinded every other error would be its own regression.

341 files / 2851 tests, all gates green.

PR #301, opened 2026-08-15
perf(tabs): bound page-extraction lifecycle

Shishir435:
Auto-refresh created one 15s interval per mounted `useTabContents()` — four call sites, several mountable at once — and each tick passed `force: true`, which bypassed the de-dup guard. So N consumers meant N concurrent full page extractions per tab (scroll to 80%, 2s whole-document MutationObserver, full re-parse).

- One refcounted timer regardless of mount count; pauses while hidden; `force` no longer bypasses the in-flight guard; unchanged content hash skips the state rewrite.
- `waitForNetworkIdle` patched `window.fetch`/`XHR.open` per caller and saved whatever it found as 'the original', so concurrent extractions left a wrapper installed forever. Now refcounted, and uninstall only restores if the current value is still ours.
- `tabContentCache` had no eviction and `clearTabContentCache` had no callers. Wired to `tabs.onRemoved`, plus LRU with entry/byte caps.

Bundle budget raised to 194k: baseline was 78 bytes under the old cap.

<h3>Greptile Summary</h3>

The PR bounds tab-content extraction and caching lifecycles while preserving independent refresh progress across selected tabs.
- Replaces per-consumer refresh intervals with one reference-counted, visibility-aware timer and per-tab extraction admission.
- Shares network-idle instrumentation safely across concurrent extraction waiters.
- Adds tab-close eviction and entry/byte limits to the extracted-content cache.
- Adjusts the Chromium background bundle budget for the resulting implementation.

The PR appears safe to merge.

No blocking failure remains; the previously reported cross-tab refresh stall is fixed by removing the sweep-wide pending lifecycle and retaining only per-tab extraction guards.

| Filename | Overview |
|----------|----------|
| src/features/tabs/hooks/use-tab-contents.ts | Introduces one visibility-aware shared refresh subscription and uses per-tab loading guards so a pending extraction no longer blocks healthy selected tabs. |
| src/lib/content-extractor.ts | Refcounts shared fetch/XHR instrumentation and conditionally restores wrappers after the final network-idle waiter settles. |
| src/lib/tools/internal/tab-utils.ts | Adds LRU entry and byte bounds while preserving signature-based cache validation and forced-read invalidation. |
| src/background/tab-lifecycle.ts | Registers guarded tab-removal cleanup for cached extracted content. |
| tools/check-bundle-size.ts | Raises the Chromium background gzip budget from 193 KB to 194 KB. |

</details>

```mermaid
flowchart TD
  C[Mounted useTabContents consumers] --> R[Reference-counted shared timer]
  V[Visibility change] --> R
  R --> S[Read current selected tab IDs]
  S --> A{Extraction already loading for tab?}
  A -- Yes --> K[Skip only that tab]
  A -- No --> E[Extract tab content]
  E --> H{Content hash and title unchanged?}
  H -- Yes --> P[Preserve cached state object]
  H -- No --> U[Update tab content state]
  T[Tab removed] --> X[Evict background cache entry]
  B[Cache entry or byte cap exceeded] --> L[Evict least-recently-used entries]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(tabs): scope auto-refresh overrun gu..."](https://github.com/shishir435/ollama-client/commit/ee66f4504b71c10e65a5293202a60274ac601a23) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53607944)</sub>

**Context used:**

- Knowledge Base — [Tool-Calling and Browser Context Access](https://app.greptile.com/shishir435/-/custom-context/knowledge-base/shishir435/ollama-client/-/docs/tools-context.md)

### DIFF
--- a/src/background/__tests__/tab-lifecycle.test.ts
+++ b/src/background/__tests__/tab-lifecycle.test.ts
@@ -1,0 +1,51 @@
+import { beforeEach, describe, expect, it, vi } from "vitest"
+
+const { listeners, mockBrowser } = vi.hoisted(() => {
+  const listeners: Array<(tabId: number) => void> = []
+  return {
+    listeners,
+    mockBrowser: {
+      tabs: {
+        onRemoved: {
+          addListener: (fn: (tabId: number) => void) => {
+            listeners.push(fn)
+          }
+        }
+      }
+    } as { tabs?: { onRemoved: { addListener: (fn: unknown) => void } } }
+  }
+})
+
+vi.mock("@/lib/browser-api", () => ({ browser: mockBrowser }))
+vi.mock("@/lib/tools/internal/tab-utils", () => ({
+  clearTabContentCache: vi.fn()
+}))
+
+import { clearTabContentCache } from "@/lib/tools/internal/tab-utils"
+import { registerTabLifecycle } from "../tab-lifecycle"
+
+describe("registerTabLifecycle", () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    listeners.length = 0
+  })
+
+  it("evicts a closed tab's cached page content", () => {
+    registerTabLifecycle()
+
+    expect(listeners).toHaveLength(1)
+    listeners[0](42)
+
+    expect(clearTabContentCache).toHaveBeenCalledWith(42)
+  })
+
+  it("does nothing on a runtime without tabs.onRemoved", () => {
+    const original = mockBrowser.tabs
+    mockBrowser.tabs = undefined
+
+    expect(() => registerTabLifecycle()).not.toThrow()
+    expect(listeners).toHaveLength(0)
+
+    mockBrowser.tabs = original
+  })
+})
--- a/src/background/index.ts
+++ b/src/background/index.ts
@@ -4,6 +4,7 @@ import { registerMessageRouter } from "@/background/message-router"
 import { startPersistenceTopology } from "@/background/persistence-readiness"
 import { registerPortRouter } from "@/background/port-router"
 import { initializeBackgroundStartup } from "@/background/startup"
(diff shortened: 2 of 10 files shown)


Shishir435:
Fixed in `ee66f450`.

Greptile is right. Removed the sweep-wide `autoRefreshInFlight` flag entirely rather than adding a timeout to it — `fetchTabContent` already refuses a second read while a tab has one in flight (`loadingIds[tabId]`, which `force` cannot bypass), and per-tab is the only scope where stacking actually matters since extractions are independent. A hung tab now blocks only itself.

Worth recording how the regression test went, because my first two attempts were green against the buggy code:

1. Asserting "healthy tab got refreshed after some ticks" passed with the flag — the *first* sweep alone satisfied it.
2. Tightening it to require progress on a later tick still passed — because the stuck extraction started at **mount**, so the sweep's call to that tab returned early on the per-tab guard and `Promise.all` resolved normally.

The wedge only reproduces when the hang begins *inside* a sweep. The committed test lets the mount read complete and hangs the first sweep read; it fails against `033081ba` and passes now.

---

Separately, the CI failure on `chat-durability.smoke.test.ts` looks unrelated to this branch and I could not reproduce it:

- 3 isolated runs and 3 full-suite runs locally, all green (367 files / 3059 tests).
- Nothing here imports a persistence path. The new `tab-lifecycle.test.ts` mocks `browser-api` and `tab-utils` and imports only `tab-lifecycle.ts`; no test in the repo imports `background/index`.
- The failing assertion is a legacy-blob write/reload round-trip, and CI ran ~5x slower than local (55s vs 11s), which is the kind of gap that surfaces timing-sensitive persistence tests.

I have not diagnosed it, so I am not claiming it is fixed. If the re-run after this push fails again it needs its own investigation rather than being attributed here.

PR #302, opened 2026-08-15
perf: throttle streaming markdown parses and narrow session load

Shishir435:
- Every token re-parsed the whole message, re-sanitized it, replaced the subtree via `innerHTML` and rebuilt every code-block toolbar. Work is O(message length) and content grows monotonically, so the last tokens of a long answer cost the most. Now throttled to 120ms; the trailing timer is anchored on the last render, not the last change, so a stream that stops mid-window still renders its final content exactly once.
- `MarkdownIt` (nine plugins) was constructed per mounted renderer. Hoisted to module scope — `render` carries nothing between calls, so footnote numbering still restarts per message.
- Session load ran `SELECT * FROM messages WHERE sessionId = ?` with no LIMIT to build a sibling map, then kept 50. Now a narrow `id/parentId/timestamp` query for the tree plus an id-restricted full read for the path. `navigateToNode` had the same full read and is fixed too.

<h3>Greptile Summary</h3>

The PR reduces streaming-render and session-loading work while completing the requested persistence-safety fixes.
- Throttles Markdown parsing while preserving a final trailing render and shares one configured parser.
- Loads narrow message-tree projections before fetching full rows only for the visible path.
- Validates every projected durable row and aborts session loading or navigation when any row is unreadable.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/lib/repositories/sqlite-chat-history.ts | Adds validated narrow tree reads and rejects partial projections, resolving both prior persistence-decoding findings. |
| src/features/sessions/stores/chat-session-message-actions.ts | Uses narrow tree reads, fetches full rows only for the active path, and preserves branch state when projection decoding fails. |
| src/features/sessions/lib/message-tree.ts | Generalizes tree helpers to operate on narrow node projections without changing traversal semantics. |
| src/hooks/use-markdown-parser.ts | Shares the MarkdownIt configuration and throttles streaming parses while retaining sanitization and trailing updates. |
| src/features/sessions/stores/__tests__/chat-session-store-actions.test.ts | Covers narrow reads, decode-failure preservation, branch navigation, pagination, and fallback behavior. |
| src/lib/repositories/__tests__/sqlite-chat-history.test.ts | Verifies projection SQL, strict decoding, safe aggregate errors, and id-restricted reads. |

</details>

```mermaid
sequenceDiagram
  participant UI as Session Store
  participant Repo as Chat History Repository
  participant DB as SQLite
  UI->>Repo: getMessageTreeBySession(sessionId)
  Repo->>DB: SELECT id, parentId, timestamp
  DB-->>Repo: narrow rows
  Repo->>Repo: Decode every tree row
  alt Any row is unreadable
    Repo-->>UI: Throw controlled decode error
    UI->>UI: Preserve existing branch state
  else Tree is valid
    Repo-->>UI: Decoded tree
    UI->>UI: Resolve paginated active path
    UI->>Repo: getMessagesByIds(pathIds)
    Repo->>DB: SELECT full rows WHERE id IN (...)
    DB-->>UI: Visible path rows
  end
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(persistence): raise on an undecodabl..."](https://github.com/shishir435/ollama-client/commit/23a618dc84765e39e79e3b5cbd24ecb082db129d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53607958)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Sessions, Memory, and Prompt Templates](https://app.greptile.com/shishir435/-/custom-context/knowledge-base/shishir435/ollama-client/-/docs/sessions-memory.md)
- Knowledge Base — [Shared State Stores and Hooks](https://app.greptile.com/shishir435/-/custom-context/knowledge-base/shishir435/ollama-client/-/docs/stores-state.md)
</details>

### DIFF
--- a/src/components/__tests__/markdown-renderer.test.tsx
+++ b/src/components/__tests__/markdown-renderer.test.tsx
@@ -446,4 +446,28 @@ describe("MarkdownRenderer", () => {
       }
     })
   })
+
+  describe("streamed content", () => {
+    it("settles on the final content instead of an intermediate parse", () => {
+      vi.useFakeTimers()
+      try {
+        const { container, rerender } = render(
+          <MarkdownRenderer content={"Streaming"} />
+        )
+
+        rerender(<MarkdownRenderer content={"Streaming ans"} />)
+        rerender(<MarkdownRenderer content={"Streaming answ"} />)
+        rerender(<MarkdownRenderer content={"Streaming **answer**."} />)
+        act(() => {
+          vi.advanceTimersByTime(500)
+        })
+
+        const surface = container.querySelector(".markdown-container")
+        expect(surface).toHaveTextContent("Streaming answer.")
+        expect(surface?.querySelector("strong")).toHaveTextContent("answer")
+      } finally {
+        vi.useRealTimers()
+      }
+    })
+  })
 })
--- a/src/features/sessions/lib/message-tree.ts
+++ b/src/features/sessions/lib/message-tree.ts
@@ -12,8 +12,23 @@ import type { ChatMessage, FileAttachment, ImageAttachment } from "@/types"
  * without touching the database.
  */
 
+/**
+ * The columns tree traversal actually reads. Every helper below is generic
+ * over this so the active path can be resolved from a narrow projection —
+ * `getMessageTreeBySession` — without loading message bodies for rows that
+ * pagination is about to discard.
+ */
+export interface MessageTreeNode {
+  id?: number | string
+  parentId?: number | string
+  timestamp?: number
+}
+
 /** Stable ordering for sibling lists. */
-export const compareMessages = (a: ChatMessage, b: ChatMessage): number => {
+export const compareMessages = (
+  a: MessageTreeNode,
+  b: MessageTreeNode
+): number => {
   const tsA = a.timestamp ?? 0
   const tsB = b.timestamp ?? 0
   if (tsA !== tsB) return tsA - tsB
@@ -25,10 +40,10 @@ export const compareMessages = (a: ChatMessage, b: ChatMessage): number => {
  * messages. Root-level messages live under the key `"root"`. Sibling
  * lists are sorted by `compareMessages`.
  */
-export const buildSiblingsMap = (
-  messages: ChatMessage[]
-): Map<number | string, ChatMessage[]> => {
-  const map = new Map<number | string, ChatMessage[]>()
+export const buildSiblingsMap = <T extends MessageTreeNode>(
+  messages: T[]
+): Map<number | string, T[]> => {
+  const map = new Map<number | string, T[]>()
   for (const msg of messages) {
     const key = msg.parentId ?? "root"
     const list = map.get(key) || []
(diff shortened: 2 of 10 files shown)


Shishir435:
Fixed in `8b26e0cd`.

`getMessageTreeBySession` now declares a Zod row schema and goes through `decodeRows`, with `table: "messages"` from the shared `DurableTable` union.

The reason it is worth doing here specifically: those three columns *are* the ancestry. A bad `id` or a non-numeric `timestamp` does not render wrong — it reorders siblings or detaches a subtree, and the visible branch just ends early with nothing to point at. That is exactly the failure mode the decode rule exists for.

Failure policy is drop-and-log, matching the ingestion/model-pull repositories rather than `tool_loop_runs`: one unreadable row costs its descendants their place in the walked path, not the whole conversation.

Two tests added; both verified failing against `2b2f93ef` — a null `id` mid-tree, and a string `timestamp`.

One thing I did **not** change, so it is a deliberate gap rather than an oversight: `messageFromRow` and `fileFromRow` in the same module still use per-field `as` assertions, and so do the session mappers. That predates this PR and is why the new function was written that way. Converting them touches every chat-history read path and belongs in its own change — happy to open an issue if you want it tracked.

Shishir435:
Fixed in `6ce9232d`. Greptile is right, and it invalidates the reasoning in my previous comment.

I justified drop-and-log by pointing at the ingestion/model-pull repositories. That analogy does not hold: **those rows are independent, tree rows are not.** Dropping one node invalidates ancestry for everything beneath it, so "costs its descendants" was understating it — dropping the node the leaf points at costs the *entire* visible conversation.

Confirmed the exact mechanism: `traversePathFromLeaf` does `if (!msg) break`, so an unresolvable leaf returns `path: []` with `hasMore: true`. The chat renders empty, with a load-more affordance, while every row sits intact in SQLite.

The fix is at the caller rather than the decoder, because the decoder cannot know which row the leaf points at. `loadSessionMessages` now resolves the stored leaf against the tree before walking, falling back to the newest node — the same fallback a session with no stored leaf already uses. It logs when they disagree, and does **not** write the repair back: a read must not repair durable state, or a transient decode failure would permanently move the user's branch.

Two things worth noting:

- **This was reachable before this PR.** Any stale `currentLeafId` — anything that removed messages without repairing it — produced the same empty render. The decoder made an existing latent bug easier to hit; the guard closes both.
- **Mid-ancestor drops stay recoverable.** The path truncates at the gap instead of going empty, and `hasMore` stays true; `loadMoreMessages` walks via `repo.getMessage(id)`, which reads full rows and never goes through the projection decoder, so it can continue past the bad node.

Two tests added. The leaf-fallback one is verified failing against `8b26e0cd`.

Shishir435:
Fixed in `23a618dc`. Both points were right, and the second one is worse than Greptile's summary states.

**The durable one.** `addMessage` reads `session.currentLeafId` from the store as the new message's `parentId` and persists it. So my "does not write the repair back" was true literally and false in effect — the guessed leaf became the parent of the next thing the user sent. `navigateToNode` was worse still: it derives a leaf from the tree and writes it with `repo.updateSession` directly, so a partial tree there moved the branch durably with no send required.

**The policy.** Two rounds of this have been me patching the symptom. Dropping rows from an interdependent structure produced a truncated branch, then a guessed leaf, then a durable re-parent — each patch creating the next hole. `getMessageTreeBySession` now **raises** when any row fails to decode. Unlike the ingestion and model-pull rows the drop policy was borrowed from, these are one structure: a partial tree is a wrong tree, not a smaller one, and the walk stopping at the first missing id is what turns that into silent data loss.

Callers no longer guess:
- `loadSessionMessages` logs and returns, leaving the session exactly as it was — including `currentLeafId`, so the next send still parents to the real branch.
- `navigateToNode` aborts before `updateSession`.

The stale-`currentLeafId` fallback from `6ce9232d` stays, and is now sound rather than a guess: reaching it means the tree decoded completely, so the newest node is a real leaf and is safe to hand `addMessage`. That case is the pre-existing one — a pointer left behind by something that removed messages without repairing it — and it used to render an empty chat.

The error text carries only counts (`1 of 5 rows unreadable`); `decodeRows` has already logged paths and codes, never values.

Five tests. The two that matter — session untouched on decode failure, and `navigateToNode` not persisting — are verified failing against `6ce9232d`. I also removed the earlier test that asserted truncation was acceptable, since that is no longer the contract.

PR #353, opened 2026-09-02
feat(docs): make the site answer agent requests correctly

Shishir435:
Fixes the Is Agentic audit failures on www.ollamaclient.in.

**Root cause:** Vercel evaluates `rewrites` after the filesystem, so the `Accept`-conditioned rewrite could never fire for a documentation page — every one is a built file. Its sibling header rule still stamped `Content-Type: text/markdown` on the HTML body it failed to replace. A static Astro build also discards the headers an endpoint module sets, which is why `/api` served `application/octet-stream` with no rate-limit fields and no way to answer an unknown path with JSON.

**Fix:** request-time behaviour moves to Vercel Routing Middleware (`docs/proxy.ts`), which runs before the filesystem. Every decision lives in `docs/src/lib/agent-routing.ts` so it is testable without a deployment; the middleware is glue and falls through to static serving on error.

- Markdown negotiation at the same URL, q-values honoured, `Vary: Accept`, `406` when nothing matches
- 404s keep a real status and gain a Markdown recovery map, shared with `/404.md`
- JSON errors for `/api/*` (404 / 405 / 406) with rate-limit and `Link` headers; `vercel.json` restates the header set for the two prerendered paths, and a test asserts it matches `apiHeaders()` exactly
- New: `/index.md`, `/.well-known/api-catalog` (RFC 9727)
- OpenAPI: typed input on all six operations (the `Origin` header the proxy really inspects), `429` + rate-limit response headers, `info.x-api-lifecycle`
- Documented Deprecation (RFC 9745) / Sunset (RFC 8594) policy with a 30-day floor
- Brand: `Organization` + `WebSite` JSON-LD with both store listings; official listings named on the About page
- `pnpm verify:agent-endpoints [baseUrl]` probes a real deployment for the whole contract

**Verification:** 40/40 endpoint checks against the built site through a local emulation of Vercel's routing order; `pnpm check:static` and 3579 tests pass.

**Needs a decision, not code:** publishing `olc` to npm/Homebrew (credentials), and press/listing work for brand-name search.

<h3>Greptile Summary</h3>

The PR adds request-time agent routing for the static documentation site and expands its machine-readable discovery surface.
- Negotiates HTML and Markdown documentation representations before static-file serving.
- Returns structured API and documentation errors while preserving correct status codes.
- Adds generated discovery resources, OpenAPI metadata, endpoint verification, and documentation for the new contracts.
- Expands site identity metadata and official distribution guidance.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| docs/src/lib/agent-routing.ts | Centralizes Accept negotiation, API routing decisions, and machine-readable error behavior; the previously reported negotiation defects are fixed. |
| docs/proxy.ts | Connects Vercel Routing Middleware to the testable routing resolver and now unconditionally matches API paths. |
| docs/src/lib/api-response.ts | Consolidates JSON response headers, error shapes, rate-limit metadata, links, and versioning policy. |
| docs/src/pages/api.ts | Expands the prerendered discovery payload and accurately documents that API endpoints ignore Accept negotiation. |
| vercel.json | Configures middleware, static API rewrites, and response headers for the generated deployment surface. |
| tools/verify/verify-agent-endpoints.ts | Adds deployment-level verification for negotiation, API errors, discovery resources, and response headers. |

</details>

```mermaid
flowchart TD
  Request[Incoming request] --> Middleware[docs/proxy.ts]
  Middleware --> Resolver[resolveRequest]
  Resolver -->|Known API GET or HEAD| StaticAPI[Prerendered JSON endpoint]
  Resolver -->|API error or OPTIONS| JSON[Structured JSON response]
  Resolver -->|Markdown accepted| Markdown[Rewrite to generated Markdown twin]
  Resolver -->|HTML accepted| StaticHTML[Static HTML serving]
  Resolver -->|Nothing acceptable| NotAcceptable[406 JSON response]
  Resolver -->|Unknown Markdown route| NotFound[404 Markdown recovery map]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["test(docs): pin the sole-exclusion Accep..."](https://github.com/shishir435/ollama-client/commit/1d5e9a4f1a4414274db5e25c9e7e1eab96a60944) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59633204)</sub>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ docs/public/llms-full.txt
 docs/public/ai.txt
 docs/public/**/*.md
 docs/public/openapi.json
+docs/public/.well-known/api-catalog
 
 # astro / starlight build caches
 docs/.astro/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -324,6 +324,44 @@ In agent mode it serves a local agent runtime over `/v1/chat/completions`, so th
 - **A browser origin is refused unless it is allowed.** The proxy listens on loopback and runs an agent, so a wildcard `Access-Control-Allow-Origin` would let any page spend a turn — a missing response header does not stop a simple request. `ALLOWED_ORIGINS` defaults to the extension schemes; a request with no `Origin` is not a page and is left alone.
 - `packages/olc/README.md` has the options, endpoints, build outputs and known limits.
 
+### Docs site and its agent surface
+
+`docs/` is an Astro static build deployed by Vercel from `vercel.json`. Two
+rules decide where a change goes.
+
+- **Vercel evaluates `rewrites` after the filesystem.** Every documentation page
+  is a built file, so nothing in `rewrites` can reshape its response. An
+  `Accept`-conditioned rewrite there is dead config; an `Accept`-conditioned
+  *header* there is worse, because it relabels a body it never replaced. That is
+  what stamped `Content-Type: text/markdown` on the HTML homepage.
+- **Request-time behaviour lives in `docs/proxy.ts`**, the Routing Middleware
+  named by `vercel.json`'s `proxy` entry, which runs before the filesystem. It is
+  an adapter only: every decision is in `docs/src/lib/agent-routing.ts` so it can
+  be tested without a deployment (`src/lib/__tests__/agent-routing.test.ts`). Its
+  `config.matcher` excludes assets, `og/`, `.well-known/`, `reference/`, and any
+  path with a file extension — which is also what stops a `.md` rewrite from
+  re-entering it.
+
+The published contract, and where each piece is owned:
+
+| Surface | Owner |
+|---|---|
+| Markdown twin of every page, negotiated at the same URL | `agent-routing.ts` + the `.md` files `generate-llms-docs.ts` writes |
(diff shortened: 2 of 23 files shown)


Shishir435:
> P1 Explicit exclusion still serves HTML — `Accept: text/html;q=0` ... `preferredRepresentation` selects HTML ... causing a 200 HTML response instead of the documented 406

This one is a false positive. The `unacceptable` check runs *before* the HTML default, so `markdown.q === 0` is only reached when HTML is actually acceptable:

```ts
if (markdown.q === 0 && html.q === 0) return "unacceptable"  // <- taken here
if (markdown.q === 0) return "html"
```

With `Accept: text/html;q=0` the sole range excludes HTML and matches nothing else, so both scores are 0 and the first line returns. Measured against the real module:

```
"text/html;q=0"          -> unacceptable   decision: respond 406
"text/html; q=0.0"       -> unacceptable   decision: respond 406
"text/*;q=0"             -> unacceptable   decision: respond 406
"text/markdown;q=0"      -> unacceptable   decision: respond 406
"text/html;q=0, */*;q=1" -> markdown       decision: rewrite
```

The concern behind it is real though — the behaviour depends on those two lines staying in that order, and nothing said so. `1d5e9a4f` pins all four cases as a test and notes the constraint at the branch.

The other three findings were correct and are fixed: range precedence (`5ea7f601`), the `/api` negotiation claim (`5ea7f601`), and the extension-path matcher (`7a008744`).

PR #355, opened 2026-09-02
feat(docs): make the site answer agent requests correctly (#353)

Shishir435:
* feat(docs): make the site answer agent requests correctly

Vercel evaluates rewrites after the filesystem, so the Accept-conditioned rewrite could never fire for a documentation page while its sibling header rule relabelled the HTML body as text/markdown. A static build also drops the headers an endpoint module sets, which is why /api served application/octet-stream with no rate-limit fields and no way to answer an unknown path with JSON.

Negotiation, agent 404s and API errors now run in Routing Middleware (docs/proxy.ts) ahead of the filesystem, with every decision in a pure module so it is testable without a deployment. Adds an RFC 9727 API catalog, a Markdown twin for the landing page, typed inputs and a rate-limit contract on every OpenAPI operation, an explicit Deprecation/Sunset policy, and brand and publisher JSON-LD.

pnpm verify:agent-endpoints probes a real deployment for the whole contract.

* fix(docs): resolve Accept quality by range precedence, not by highest q

RFC 9110 settles a media type's quality with the most specific matching range. Scoring by the highest q let a permissive wildcard override an exact exclusion, so `text/html;q=0, */*` served the HTML the client had just ruled out.

Also stops the new contract tests reading generated artifacts under docs/public, which is gitignored: they passed locally and failed on a clean CI checkout. They assert on the generator's builders instead, which is the thing the artifact is written from.

Scopes the negotiation contract in /api and the portal to documentation pages; /api is JSON only and ignores Accept.

* fix(docs): match /api unconditionally in the routing middleware

The matcher drops any path with a file extension, which is what stops a .md rewrite re-entering the middleware. It also dropped /api/models.json, handing a machine the static HTML 404 under the one prefix that promises JSON. /api is now matched beside that rule rather than through it.

* test(docs): pin the sole-exclusion Accept case at 406

A review read the HTML default as reachable for `Accept: text/html;q=0`. It is not — the unacceptable check precedes it — but nothing said so, and swapping the two lines would silently answer 200 HTML to a client that excluded HTML. Now a test holds the order.

<h3>Greptile Summary</h3>

The follow-up centralizes legacy documentation aliases so Astro and the routing middleware use the same redirect definitions, and it corrects the Accept-negotiation documentation.
- Markdown requests for legacy URLs now receive canonical redirects before filesystem routing.
- Shared redirect tests and deployment verification cover the alias contract.
- The previously flagged standalone prose was moved into JSDoc.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| docs/src/lib/agent-routing.ts | Uses the shared legacy redirect map before Markdown page lookup and replaces the previously flagged standalone prose with JSDoc. |
| docs/src/seo/legacy-redirects.mjs | Defines the three existing legacy aliases once for both Astro and routing middleware. |
| docs/astro.config.mjs | Replaces the inline redirect object with the shared legacy alias map. |
| docs/proxy.ts | Supplies the shared alias map to the pure routing decision module. |
| src/lib/__tests__/agent-routing.test.ts | Verifies canonical Markdown redirects, trailing-slash handling, shared-map wiring, and Accept exclusion behavior. |

</details>

```mermaid
flowchart TD
  R[Incoming documentation request] --> M[Routing Middleware]
  M --> A{Markdown preferred?}
  A -- No --> F[Static filesystem serving]
  A -- Yes --> L{Legacy alias?}
  L -- Yes --> C[308 canonical redirect]
  L -- No --> D{Published Markdown twin?}
  D -- Yes --> W[Rewrite to generated .md file]
  D -- No --> N[Markdown 404 response]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(docs): answer legacy inbound URLs fo..."](https://github.com/shishir435/ollama-client/commit/e8019de6160522ebf4dd7b54b81d8f73d8df076c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59647625)</sub>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ docs/public/llms-full.txt
 docs/public/ai.txt
 docs/public/**/*.md
 docs/public/openapi.json
+docs/public/.well-known/api-catalog
 
 # astro / starlight build caches
 docs/.astro/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -324,6 +324,44 @@ In agent mode it serves a local agent runtime over `/v1/chat/completions`, so th
 - **A browser origin is refused unless it is allowed.** The proxy listens on loopback and runs an agent, so a wildcard `Access-Control-Allow-Origin` would let any page spend a turn — a missing response header does not stop a simple request. `ALLOWED_ORIGINS` defaults to the extension schemes; a request with no `Origin` is not a page and is left alone.
 - `packages/olc/README.md` has the options, endpoints, build outputs and known limits.
 
+### Docs site and its agent surface
+
+`docs/` is an Astro static build deployed by Vercel from `vercel.json`. Two
+rules decide where a change goes.
+
+- **Vercel evaluates `rewrites` after the filesystem.** Every documentation page
+  is a built file, so nothing in `rewrites` can reshape its response. An
+  `Accept`-conditioned rewrite there is dead config; an `Accept`-conditioned
+  *header* there is worse, because it relabels a body it never replaced. That is
+  what stamped `Content-Type: text/markdown` on the HTML homepage.
+- **Request-time behaviour lives in `docs/proxy.ts`**, the Routing Middleware
+  named by `vercel.json`'s `proxy` entry, which runs before the filesystem. It is
+  an adapter only: every decision is in `docs/src/lib/agent-routing.ts` so it can
+  be tested without a deployment (`src/lib/__tests__/agent-routing.test.ts`). Its
+  `config.matcher` excludes assets, `og/`, `.well-known/`, `reference/`, and any
+  path with a file extension — which is also what stops a `.md` rewrite from
+  re-entering it.
+
+The published contract, and where each piece is owned:
+
+| Surface | Owner |
+|---|---|
+| Markdown twin of every page, negotiated at the same URL | `agent-routing.ts` + the `.md` files `generate-llms-docs.ts` writes |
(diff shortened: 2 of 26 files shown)


Shishir435:
Both findings checked. One is real and fixed in #356 (into `preview`, so this PR picks it up on the next sync).

**P1 Redirect aliases become Markdown 404s — correct, and a genuine regression.** Reproduced against the module:

```
/architecture          text/markdown  -> respond 404
/architecture          text/html      -> pass
/ollama-setup-guide    text/markdown  -> respond 404
/privacy-policy        text/markdown  -> respond 404
```

The middleware runs before the filesystem and knew only `DOC_ORDER`, so it never saw the meta-refresh pages Astro emits. All three paths are still linked from the store listing, the README and existing search results. Fixed by moving the map to `docs/src/seo/legacy-redirects.mjs`, feeding both the Astro config and the middleware from it, and answering `308` to the canonical path for Markdown clients. Browsers keep the meta-refresh page. `verify:agent-endpoints` now probes every entry in the map.

**P2 Prose comment violates documentation contract — does not apply.** `config/__tests__/documentation-comments.test.ts` scans `src`, `packages`, `config` and `tools`; `docs/` is not a source root. Its pattern is also `/^\/\/(?!\/)/`, anchored at column 0, and that comment is indented inside a function body. Static checks pass on the commit for both reasons.

The underlying point was fair, though — that prose is about the function's algorithm, so it now lives in `preferredRepresentation`'s JSDoc rather than as a floating comment.

PR #367, opened 2026-09-08
fix(agent): observe pages the builder used to reject

Shishir435:
Two perception defects made whole pages unusable for a run, plus the anonymous failure that hid them.

**What changed**
- Every observed element field is derived through a check that answers for a non-HTML candidate. `[role]` admits an `<svg role="img">`, whose absent `isContentEditable` left `editable` undefined and failed the entire snapshot.
- Visible candidates are selected before the 2 000-element cap, in document order. A page's hidden controls could previously consume the whole budget.
- Computed styles and visibility are memoized per pass, which is what makes resolving *every* candidate affordable.
- A bound control request the content script cannot serve is answered with a typed `agent_control_failed` carrying schema paths and rule codes only. The registry does not reopen for it, and the run says which stage failed instead of "could not be observed safely".
- The agent e2e spec is split into per-family files over a shared scenario harness (`e2e/chromium/fixtures/agent-scenario.ts`), so a scenario is one file instead of an edit to the one file every agent change touches.

**Exit evidence** (7/7 `@critical` agent scenarios pass against the production build)
| page | before | after |
|---|---|---|
| 2 100 hidden inputs + 1 button | 2 000 rows, 0 visible — button dropped, 47.3 ms | 2 000 rows, 1 visible — button kept, 27.2 ms |
| 600 sections (~3 000 candidates) | 482 KB, 205.5 ms | 482 KB, 151.5 ms |
| depth 120 × 20 subtrees | 102.7 ms | 94.0 ms |

New e2e scenarios `svg` and `starved` assert both fixes end to end. Background bundle 225 939 / 226 000 gzip — under budget, 61 bytes of headroom left, so the next agent PR will need a documented bump.

**Not in this PR**
Bounded whole-document text, structural grouping and the compact projection — all of them change what the model is told, which this PR deliberately does not.

<h3>Greptile Summary</h3>

This PR makes browser-agent observations tolerate non-HTML role-bearing elements, prioritizes visible controls within the observation cap, adds bounded and memoized document traversal, and propagates typed control failures through the content-script, background, and runtime layers.
- Splits the critical agent E2E suite into scenario-specific specifications backed by a shared harness.
- Adds end-to-end coverage for SVG candidates and controls following thousands of hidden inputs.
- Preserves structural schema diagnostics without forwarding rejected page content.
- Adds a wall-clock guard to candidate selection and truncates visible text when the shared pass budget expires.

The PR appears safe to merge because no new actionable defect remains in the scoring set.

The visible-control starvation finding is fixed by selecting candidates before applying the 2,000-element output cap. Both previous threads were manually resolved without explanatory replies. The remaining concern about coarse deadline checks overlaps the previously resolved unbounded-traversal thread and is therefore not reposted as a new finding.

| Filename | Overview |
|----------|----------|
| src/lib/browser-agent/observation-builder.ts | Adds non-HTML-safe field derivation, per-pass visibility/style memoization, visible-first candidate selection, and a shared 500 ms traversal budget. |
| src/lib/browser-agent/control-port.ts | Introduces request-bound typed failure responses containing capped schema paths and codes while keeping the content port available. |
| src/background/agent/agent-control-sessions.ts | Propagates deterministic typed content-script failures without reopening and repeating the observation. |
| packages/agent-runtime/src/control-failure.ts | Defines safe typed control failures and maps observation-stage reasons to user-visible messages. |
| e2e/chromium/fixtures/agent-scenario.ts | Extracts the critical agent scenario infrastructure into a reusable production-boundary test harness. |

</details>

```mermaid
flowchart LR
  A[Agent controller] -->|observe request| B[Background session registry]
  B --> C[Bound content-script port]
  C --> D[Observation builder]
  D --> E[Select visible-first candidates]
  E --> F[Build bounded observation]
  F -->|valid snapshot| C
  D -->|build or schema failure| G[Bound agent_control_failed response]
  G --> B
  B --> H[Typed AgentControlFailedError]
  H --> A
  A --> I[Stage-specific run failure message]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(agent): bound the observation pass b..."](https://github.com/shishir435/ollama-client/commit/66db1e0a1394a83baba867c2bf1d83289102295e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61726449)</sub>


Shishir435:
Confirmed and fixed in fc5a0ed8 — the finding is exactly right, and the fixture proves it: on the reviewed commit that page observes `[]` where the button should be.

The candidate cutoff was the starvation defect one page-size later, so bounding the scan by position was never going to work. Selection now resolves **every** candidate the selector matches, in document order, and stops only once the visible budget is full — nothing past that point could enter the set anyway. Hidden candidates are retained only up to the budget they could claim, and the surplus is dropped from the end, so the result stays in document order and within 2 000 rows.

Removing the bound costs nothing, because the visibility predicates are a conjunction and their order is free: asking for the element's own box *before* its ancestors' computed styles answers an unrendered candidate without resolving a style at all.

| fixture | before this branch | reviewed commit | now |
|---|---|---|---|
| 2 100 hidden + 1 button | 47.3 ms, **button lost** | 27.2 ms | **25.7 ms** |
| 30 000 hidden + 1 button | button lost | **button lost** | **179.4 ms, button found** |
| 600 sections (~3 000 candidates) | 205.5 ms | 151.5 ms | **147.7 ms** |

A 30 000-candidate page now finds its one button in less time than a 3 000-candidate page took before this branch. Regression test `surfaces a visible control behind more hidden ones than any cap` places the control past 20 100 hidden candidates and fails on the previous commit. `AGENT_OBSERVATION_LIMITS.candidates` is gone rather than raised.

Gate: 3 990 unit tests, typecheck, lint, both browser packages, 7/7 `@critical` agent e2e against the production build, background bundle unchanged at 225 939 / 226 000 gzip.

Shishir435:
Fixed in 66db1e0a. The pass now carries a 500ms wall-clock budget, read once per 256 candidates so it costs nothing below it.

Selection raises when the budget runs out rather than returning what it has — a truncated selection is the defect it exists to prevent. Text collection stops instead, because `visibleText` already truncates at its own cap. Scale fixtures pin the clock, so they assert coverage and not timing.

PR #368, opened 2026-09-08
fix(olc): release a request slot when its caller leaves

Shishir435:
The single-flight queue took no caller signal, so a request whose connection closed still held or claimed the slot.

**What changed**
- The queue accepts a caller `signal` and honours it per stage, because they are different problems: a request still **queued** is dropped and never starts; one already **running** is cancelled the way a deadline cancels it — task aborted, slot held until it unwinds — since a task that is still running has not left the single-flight boundary.
- Both routes bind that signal from one shared `bindRequestAbort`, **before** the queue, where a request can wait for as long as another turn is allowed to run. The image route was claiming a slot for an abandoned request too.
- Backend startup is deliberately *not* cancelled — the next request reuses it — but a request whose caller left no longer spends a turn behind it. This needed no widening of `AgentBackend`.
- `REQUEST_TIMEOUT_MS` 1 800 000 → 300 000. A slot lived `REQUEST_TIMEOUT_MS + 60_000` = 31 min while the extension abandons an agent decision after 120 s, so the old default only bought a slot that outlived every client able to hold it.
- A stream failure is emitted as an OpenAI `error` event instead of `[Proxy Error]` prose finished with `stop`, which read as a *completed* turn to every compatible client — that is how a provider error reached the extension as a decision with no tool call. `src/lib/providers/__tests__/provider-tools.test.ts` already covers the extension parsing that frame; nothing under `src/` changes here.

**Exit evidence** — 337/337 olc tests, 3 972 total
- Hold A, disconnect queued B, release A → B never starts (`startTurn` stays 1).
- Disconnect A mid-stream → the next request is admitted only after A unwinds (queue log: `Starting … after waiting 362ms`), and A's turn was aborted exactly once.
- Disconnect during `ensureReady` → `ensureReady` still ran once, `startTurn` zero times.
- Same for an abandoned image request.

**Not in this PR**
Parked-call semantics — every agent decision still parks a call that is never resumed. Recorded as a gap, not fixed here.

<h3>Greptile Summary</h3>

This PR makes the OLC single-flight queue responsive to caller disconnects while preserving the invariant that a running task retains its slot until it unwinds. It also shortens the default request timeout and represents backend stream failures as OpenAI-compatible error events.

- Drops disconnected requests before they begin executing.
- Cancels running requests when their callers leave without admitting overlapping work.
- Binds image-request cancellation before asynchronous model resolution.
- Returns structured errors for streaming and non-streaming backend failures.
- Adds focused queue, route, and HTTP lifecycle coverage.
- Fixes the prior declaration-comment rule violation by converting the timeout explanation to JSDoc.

The PR appears safe to merge, with both previous findings resolved and no new actionable defects identified.

The image route now binds disconnect handling before model resolution, and the timeout declaration now uses JSDoc; the current changes introduce no remaining blocking or non-blocking findings.

| Filename | Overview |
|----------|----------|
| packages/olc/src/core/queue.ts | Adds stage-aware caller cancellation while retaining the queue slot until running work has unwound. |
| packages/olc/src/core/http.ts | Introduces shared disconnect binding that handles both future events and connections closed before listener registration. |
| packages/olc/src/core/chat-route.ts | Threads caller cancellation through queue admission and emits structured backend failure events. |
| packages/olc/src/core/image-route.ts | Binds cancellation before model resolution and propagates it into image queue admission. |
| packages/olc/src/config.ts | Reduces the default request timeout to five minutes and uses the required JSDoc declaration prose. |

</details>

```mermaid
flowchart TD
  A[HTTP request arrives] --> B[Bind caller abort signal]
  B --> C{Caller still connected?}
  C -- No --> D[Reject without entering queue]
  C -- Yes --> E[Enter single-flight queue]
  E --> F{Caller leaves while queued?}
  F -- Yes --> G[Remove queued entry; task never starts]
  F -- No --> H[Start task with queue abort signal]
  H --> I{Caller leaves or deadline expires?}
  I -- No --> J[Complete task and release slot]
  I -- Yes --> K[Abort running task]
  K --> L{Task unwinds within grace period?}
  L -- Yes --> M[Release slot and admit next request]
  L -- No --> N[Mark queue stalled and refuse new requests]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Update packages/olc/src/config.ts"](https://github.com/shishir435/ollama-client/commit/b304647e7d33737e6f4239b3101a71d1f7c66a09) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61730365)</sub>

### DIFF
--- a/packages/olc/README.md
+++ b/packages/olc/README.md
@@ -327,6 +327,7 @@ readiness reporting. Ollama itself is installed separately.
 | `SYSTEM_PROMPT` | `--system-prompt` | `OLC_SYSTEM_PROMPT` | the client's |
 | `BRIDGE_ENABLED` | `--no-bridge` to disable | `OLC_BRIDGE_ENABLED` | `true` |
 | `DEBUG` | `--debug` | `OLC_DEBUG` | `false` |
+| `REQUEST_TIMEOUT_MS` | — | `OLC_REQUEST_TIMEOUT_MS` | `300000` |
 
 `REQUEST_TIMEOUT_MS`, `BRIDGE_CALL_TIMEOUT_MS`, `BRIDGE_BATCH_MS` and
 `SUSPENDED_TURN_TTL_MS` follow the same precedence with `OLC_`-prefixed
@@ -486,6 +487,11 @@ new adapter's expectations.
   cancelled, not merely failed, and the next request waits for it to unwind. A
   cancelled turn that does not stop is never overtaken: after ten seconds the
   queue refuses requests with `503` and names it, until it stops.
+- **A departed client releases its slot.** A request whose connection closes
+  while it is still queued is dropped and never starts; one that has already
+  started is cancelled the same way a deadline cancels it, and the slot is held
+  until it unwinds. Backend startup is not cancelled — the next request reuses
+  it — but a request whose caller left does not spend a turn behind it.
 - **A turn per user message.** Conversation history is replayed from the client's
   messages; only a tool exchange reuses its turn. Trailing tool results whose turn
   the proxy no longer holds — expired, cancelled, or from an earlier run — are
--- a/packages/olc/src/config.ts
+++ b/packages/olc/src/config.ts
@@ -30,7 +30,13 @@ export const DEFAULTS = {
     "moz-extension://*",
     "safari-web-extension://*"
   ],
-  REQUEST_TIMEOUT_MS: 1_800_000,
+  /*
+   * A caller's own deadline is shorter than this: the extension abandons an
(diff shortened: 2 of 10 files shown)


Shishir435:
Fixed in 4d00c805. The route binds the abort signal before its first await, and `bindRequestAbort` now reads the connection state as well as subscribing to it — a listener only hears what has not happened yet.

That state is read from the response only: Node destroys a fully consumed request stream by itself, so `request.raw.destroyed` is true for every ordinary POST. Using it returned 499 for live requests.

Regression test aborts the fetch during model resolution and asserts `generateImage` is never called; it fails on 516dd3f7.

PR #369, opened 2026-09-08
fix(agent): make a wrong command cost a retry, not the run

Shishir435:
Whether an observed control accepts a command was decided only by the trusted resolver, which runs after the decision is accepted — so a wrong command ended the run instead of being corrected. The rule now lives once in a pure classifier (`packages/agent-runtime/src/affordance.ts`) that the parser and the resolver both ask, and the parser asks first.

The retry is told why. It carried a counter and nothing else, which is how the same wrong answer came back until the malformed budget ran out. The feedback is templates plus the model's own ref and structural facts — never a page string, since it becomes part of a prompt.

Also: a refused command records `invalid_decision`, not `verification_failed` (nothing touched the page); a model that gives up records the new `goal_failed` instead of blaming the endpoint; and both active-time ceilings are now enforced, having been recorded in the checkpoint and never read since 0.14.0. Deletes `createAgentBudgetTracker` and `agent-dry-run.ts`, which had no callers.

Two fixture-driven calls worth flagging: clicking a checkbox used to refuse as "not an activatable control — click a button, a link or a menuitem", which a model cannot act on, so it now gets `use_check_instead`; and the resolver's single "stale or ambiguous" message covered two unrelated refusals and now names each.

Background budget 226 000 → 228 000, measured 227 325, with the comment naming what moved in.

Gate: 4 050 unit tests, 8/8 `@critical` agent e2e against the production build. No locale keys — agent error codes have no i18n coupling.

<h3>Greptile Summary</h3>

The PR moves browser-control affordance checks into a shared classifier used before accepting a model decision and again during trusted resolution, allowing invalid commands to consume a retry with actionable feedback instead of terminating the run. It also:
- Filters page-authored role and input-type values before including structural details in retry prompts.
- Distinguishes grounding, stale-page, unreadable-page, and generic resolver failures.
- Enforces durable active-time budgets and introduces a dedicated `goal_failed` result.
- Removes unused budget-tracker and dry-run code, adds unit and critical end-to-end coverage, and updates dependency overrides.

The PR appears safe to merge; both previously reported problems are resolved and no actionable new regression was established.

Corrective feedback now filters arbitrary page-authored role and type values, resolving the prompt-injection path, and resolver failures are separated into grounding, stale-page, unreadable-page, and fallback classifications rather than all being attributed to invalid model decisions. Both previous threads were manually resolved, and the current code supports those resolutions.

| Filename | Overview |
|----------|----------|
| packages/agent-runtime/src/affordance.ts | Centralizes command-to-control affordance classification and produces filtered, actionable retry feedback. |
| packages/agent-runtime/src/controller.ts | Enforces active-time budgets and maps resolver and model-declared failures to more accurate run error codes. |
| packages/agent-runtime/src/resolution-failure.ts | Introduces typed resolver failures and translates them into stable agent error codes. |
| src/application/agent/agent-decision-parser.ts | Rejects unsupported commands before acceptance so they consume the bounded model retry budget. |
| src/application/agent/agent-model-port.ts | Includes actionable parser feedback in subsequent model attempts while preserving malformed-response limits. |
| src/lib/browser-agent/resolved-effect.ts | Reuses the shared affordance classifier and throws typed failures for stale or unreadable browser state. |
| pnpm-workspace.yaml | Raises several dependency override floors without introducing an established peer or engine conflict. |
| pnpm-lock.yaml | Regenerates resolutions for the updated security and documentation dependency overrides. |

</details>

```mermaid
sequenceDiagram
  participant M as Model
  participant P as Decision parser
  participant C as Shared affordance classifier
  participant R as Trusted resolver
  participant B as Browser page

  M->>P: Proposed command
  P->>C: Classify against observation
  alt Command is unsupported
    C-->>P: Structured refusal
    P-->>M: Retry with filtered feedback
  else Command is supported
    P-->>R: Accepted decision
    R->>C: Re-check current observation
    alt Grounding refusal
      C-->>R: AgentGroundingError
    else Resolvable
      R->>B: Execute verified effect
    end
  end
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(deps): pin patched transitive adviso..."](https://github.com/shishir435/ollama-client/commit/4f09a0b744a346814496aedb76d8176af51bb21b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61771088)</sub>


Shishir435:
Both findings were already fixed in 200120b6, which is now pushed — the review above is against 8c84f80a, before it. The gate had rejected that push over unrelated production advisories (#370).

**Page text in feedback.** Right, and it broke the invariant this PR claims. `role` and `type` now survive into a refusal only if they name something a bounded vocabulary knows: 20 ARIA roles, 25 type keywords, anything else dropped and the element described by tag alone. Filtered where the refusal is built *and* inside the exported feedback builder — my first attempt only did the former, and the test going through the public function caught it.

**Resolver classification.** Also right, and a regression I introduced. `AgentStaleObservationError` → `stale_snapshot`, `AgentUnreadablePageError` → `unsupported_page`, grounding refusal → `invalid_decision`, anything unrecognized keeps `verification_failed` rather than being quietly relabelled.

4f09a0b7 is #370's dependency pin riding along so this branch can pass its own audit; it drops out on rebase once that merges.

PR #371, opened 2026-09-08
feat(agent): let a decision see what the run already did

Shishir435:
Every decision was made from the current page alone, so a run that had just filled a field saw a filled field and no reason it was filled. `AgentModelInput.previousVerification` had been declared since the loop was written and never populated once.

- A receipt records what the step acted on — role, tag, bounded name, source URL — because a command holds a ref and a ref means nothing after the next observation. The name is page text, so it is dropped for a sensitive control and bounded both where it is assembled and where it is written.
- History is built from those receipts, not from memory, so a worker restart keeps it; a read that fails degrades the decision rather than ending the run.
- Outcome comes from the verification, never from the fact that a step was attempted: `executed` without a confirmation reports as unverified. Bounded by 12 steps and ~6 KB, dropped oldest first so the same state always renders the same way.
- D5 taken as proposed: an optional bounded `finding` on a command decision, written on the receipt for the decision that made it.
- G13: the navigation fixture now asserts arrival and origin instead of `click` vs `navigate`.

Two new `@critical` scenarios: `continuity` asserts request 2 contains action 1 and its confirmed outcome, and `compare` answers from two pages where the second cannot show the first fact.

Gate: 4 087 unit tests, 10/10 agent e2e against the production build. Background budget 228 000 → 229 500, measured 228 969, comment names what moved in.

The harness now reports `run.error` and the step list when a status poll times out — that is how the looping `compare` fixture was diagnosed, rather than "expected completed, received failed".

<h3>Greptile Summary</h3>

This PR gives each agent decision a bounded, durable account of earlier steps and their verified outcomes.
- Records sanitized source-page identity, target metadata, and optional model findings in step receipts.
- Reconstructs deterministic history after worker restarts and supplies it to subsequent model decisions.
- Bounds history by step count and encoded size while distinguishing confirmed actions from unverified attempts.
- Adds continuity and cross-page comparison scenarios, repository validation, and improved diagnostic tracing.
- The three previous findings are addressed: URL secrets are stripped, oversized lone entries are reduced, and history-read failures are traced.

The PR appears safe to merge; no blocking or independently actionable new issue remains.

The current changes sanitize durable URLs, enforce the history bound for lone entries, and expose history-read degradation through structural tracing, fully addressing the previous findings. The new receipt-to-history-to-model flow remains bounded and preserves verification semantics.

| Filename | Overview |
|----------|----------|
| packages/agent-runtime/src/controller.ts | Reads durable receipts before each decision, supplies reconstructed history, records bounded step evidence, and traces degraded history reads. |
| packages/agent-runtime/src/history.ts | Builds deterministic, size-bounded history while sanitizing URLs and deriving outcomes from verification. |
| src/lib/repositories/agent-runs.ts | Persists and validates the new bounded target, URL, and finding receipt fields and exposes ordered step reads. |
| src/application/agent/agent-model-port.ts | Adds history and previous-step outcome to the model request while preserving bounded malformed-response retries. |
| packages/contracts/src/agent.ts | Extends command decisions with an optional bounded finding. |
| packages/contracts/src/agent-panel.ts | Extends panel-visible step records with validated target, source URL, and finding fields. |
| e2e/chromium/critical/__tests__/agent-continuity.spec.ts | Verifies continuity across decisions and synthesis of facts collected from separate pages. |

</details>

```mermaid
sequenceDiagram
    participant C as Agent Controller
    participant P as Step Persistence
    participant H as History Builder
    participant M as Agent Model
    C->>P: Read durable step receipts
    P-->>C: Ordered lifecycle receipts
    C->>H: Build bounded history
    H-->>C: Steps, verified outcomes, targets, sanitized URLs
    C->>M: Current observation + prior history
    M-->>C: Command and optional finding
    C->>P: Append planned/executed/verified receipts
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(agent): reduce a step&#39;s URL to the p..."](https://github.com/shishir435/ollama-client/commit/07523cebe4e93e860625e366ce40880cdcc1198c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61854360)</sub>

**Context used:**

- Knowledge Base — [Durable data and recovery](https://app.greptile.com/shishir435/-/custom-context/knowledge-base/shishir435/ollama-client/-/docs/durable-data.md)


Shishir435:
Fixed in 07523ceb.

**URL secrets.** Correct, and it was a privacy defect: `location.href` went whole into a durable receipt and then back out to the panel and the next prompt. Receipts now keep origin and path only — what distinguishes one page from another — stripped on the write and again on the read, and a navigate destination is reduced before it is described. `https://user:pass@Example.com/orders?token=abc#at=xyz` becomes `https://example.com/orders`. Parsed structurally rather than with `URL`, since this package has no DOM lib by design.

**Byte bound.** Also correct — the `length > 1` guard let a lone oversized entry through, which is the case that empties a small local model's context. Per-field caps make it unlikely; a final reduction dropping finding, evidence, target then url makes it impossible. A 400-char finding under an 80-byte bound now returns `{step, action, outcome}`.

**Silent degradation.** A run that lost continuity is indistinguishable from a model behaving badly. Added an optional `trace` on the controller dependencies, structural metadata only, wired to `traceAgentRun`; the catch reports `history_unavailable` and still degrades rather than failing the run.

4 102 tests, 10/10 agent e2e.

PR #372, opened 2026-09-08
feat(agent): send the model a page, not the executor's bookkeeping

Shishir435:
The whole observation was serialized into every decision prompt, and most of it is not the model's business: frame ids that are always zero, verification bindings, form fingerprints, flags already at their default.

| page | raw | projected |
|---|---|---|
| form | 949 B (~271 tok) | **334 B (~95 tok)** |
| app-like, 1 202 controls | 334 KB (~95 k tok) | **140 KB (~40 k tok)** |
| text-heavy article | 22.6 KB | 22.2 KB — correct, its payload *is* the text |

The projection keeps the observation's shape rather than inventing a text format, so refs stay authoritative and the e2e fixtures stay readable; it states only the unusual side of each flag.

- **`num_ctx` is now set.** Nothing set one, so Ollama applied its own default and dropped everything past it from the front — where the system prompt and tool schema are. That surfaced as a malformed decision, never as a context error. Sized from the request, clamped 8 192–32 768.
- **Open dialogs and menus are reported** (G8/G16), in a new `modals` field — not `dialogs`, which is typed for native alert/confirm/prompt and stays empty because those block the page and cannot be read.
- **Every element names its landmark, form or dialog.** Two controls with the same accessible name are ordinary; a flat list gave a decision nothing to tell them apart with.
- **Bounded whole-document text**, moved here from PR 1 because it changes the prompt. Sent only when it says more than the viewport did, and it declares when it was cut.

New `@critical` `modal` scenario: opener then the dialog's own control, asserting two verified steps rather than the opener twice.

**Known limit worth a threshold in PR 8:** an app-like page still projects to ~40 k tokens, past the 32 768 ceiling. The element cap is the real bound, and lowering it would reintroduce the starvation PR 1 fixed — so this is measured and stated, not papered over.

Gate: 4 123 unit tests, 11/11 agent e2e. Background budget 229 500 → 230 500, measured 229 855.

Also fixed en route: `form.name` returns the input named "name", not the attribute — form-property shadowing made a group label an element instead of a string and threw, which the `form` fixture caught as an unobservable page.

**Not in this PR:** the tab inventory. It is a new port with a privacy decision attached (D2 was proposed, never confirmed), so it wants its own review rather than riding along.

<h3>Greptile Summary</h3>

This PR reduces agent prompt size by projecting observations into a model-facing page representation while preserving exact actionable values. It also sizes Ollama context windows, reports modal ownership and document-wide text, and aligns navigation data-egress analysis with all text exposed to the model.
- Removes executor-only observation bookkeeping and default-valued flags from prompts.
- Preserves complete enabled select options and exact field values.
- Adds modal and structural grouping context for duplicate controls.
- Adds bounded document-wide text with explicit truncation reporting.
- Sets a request-derived Ollama context window between 8,192 and 32,768 tokens.
- Extends copied-page-data detection to off-viewport text.

The PR appears safe to merge; the changes since the previous review fully address the outstanding observation-truncation and data-egress findings without introducing a confirmed new defect.

Budget exhaustion now explicitly marks document text as truncated, and copied-page-data analysis now includes the document text exposed to the model. Complete select values and all enabled options are preserved; that earlier thread was manually resolved without explanation.

| Filename | Overview |
|----------|----------|
| src/application/agent/agent-observation-projection.ts | Projects observations into a smaller model-facing representation while retaining complete actionable values. |
| src/application/agent/agent-model-port.ts | Sends projected observations and requests a bounded context window derived from prompt size. |
| src/lib/browser-agent/observation-builder.ts | Adds modal ownership, structural element groups, and bounded document-wide text with truncation metadata. |
| src/lib/browser-agent/resolved-effect.ts | Includes off-viewport document text when detecting page data copied into model-composed destinations. |
| packages/contracts/src/agent-observation.ts | Extends observation contracts with element groups, modal state, and bounded document text. |
| tools/checks/check-bundle-size.ts | Raises the Chromium background bundle budget to cover the measured cost of the new observation features. |

</details>

```mermaid
flowchart LR
  Page[Browser page] --> Builder[Observation builder]
  Builder --> Full[Executor observation]
  Full --> Executor[Grounding and verification]
  Full --> Projection[Model-facing projection]
  Projection --> Prompt[Decision prompt]
  Prompt --> Model[Ollama model]
  Model --> Decision[Agent decision]
  Decision --> Policy[Effect and egress policy]
  Policy --> Executor
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(agent): keep a select&#39;s options exac..."](https://github.com/shishir435/ollama-client/commit/df863a9f45425bfbb4fed30f6e09e66e3e89adb5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61866860)</sub>

### DIFF
--- a/e2e/chromium/critical/__tests__/agent-click.spec.ts
+++ b/e2e/chromium/critical/__tests__/agent-click.spec.ts
@@ -14,7 +14,7 @@ const silentButton =
   '<button type="button" onclick="fetch(\'/effect\');">Continue</button>'
 
 const clickContinue = (observation: AgentFixtureObservation) =>
-  observation.visibleText.includes("Status: Active")
+  observation.text.includes("Status: Active")
     ? { type: "complete", summary: "Active" }
     : {
         type: "click",
--- a/e2e/chromium/critical/__tests__/agent-continuity.spec.ts
+++ b/e2e/chromium/critical/__tests__/agent-continuity.spec.ts
@@ -25,7 +25,7 @@ runAgentScenario({
   html: () =>
     `<!doctype html><title>Agent continuity</title><main>${agentConfirmingButton()}</main>`,
   decide: (observation: AgentFixtureObservation) =>
-    observation.visibleText.includes("Status: Active")
+    observation.text.includes("Status: Active")
       ? { type: "complete", summary: "Active" }
       : {
           type: "click",
--- a/e2e/chromium/critical/__tests__/agent-form.spec.ts
+++ b/e2e/chromium/critical/__tests__/agent-form.spec.ts
@@ -15,7 +15,7 @@ runAgentScenario({
   hosted: true,
   html: (path) => (path.startsWith("/details") ? AGENT_DETAILS_PAGE : formPage),
   decide(observation) {
-    if (observation.visibleText.includes("Status: Active"))
+    if (observation.text.includes("Status: Active"))
       return { type: "complete", summary: "Active" }
     const field = agentFixtureElement(
       observation,
--- a/e2e/chromium/critical/__tests__/agent-grounding.spec.ts
+++ b/e2e/chromium/critical/__tests__/agent-grounding.spec.ts
@@ -18,7 +18,7 @@ runAgentScenario({
   html: () =>
     '<!doctype html><title>Agent grounding</title><main><h1>Preferences</h1><label for="news">Newsletter</label><input id="news" type="checkbox" onchange="document.querySelector(\'main\').insertAdjacentHTML(\'beforeend\',\'<p>Status: Active</p>\')"></main>',
   decide(observation: AgentFixtureObservation, { step }) {
-    if (observation.visibleText.includes("Status: Active"))
+    if (observation.text.includes("Status: Active"))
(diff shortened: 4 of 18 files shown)


Shishir435:
Fixed in df863a9f.

**Select values.** Right, and it was a functional break: the executor requires exact equality against the option the page holds, so a truncated value was unselectable and option 26 unreachable. Every enabled option now travels verbatim, with a label only when it says something the value does not.

I found the same defect one field over — `value` was capped at 200 chars, and for a select that has to match an option exactly. Cap removed there too. `name` keeps its cap: prose the model reads and never reproduces.

**Budget truncation.** Correct. The walk broke on budget exhaustion with `truncated` still false, so a complex page could send partial text that looked complete. Both exits now declare it, with a stalled-clock test.

**Off-viewport egress.** Correct and the sharpest of the three. Destination evidence now reads `documentText` alongside `visibleText`, with a test that words caught on screen are equally caught below the fold.

4 127 tests, 11/11 agent e2e.

PR #373, opened 2026-09-08
feat(agent): scope an approval to a run, and answer its questions

Shishir435:
Filling three fields cost three prompts, which is what trains a user to approve without reading — the failure a confirmation exists to prevent. An approval can now be widened to the origin it names, for the rest of that run only.

**What may be widened is policy's answer, not the panel's.** The approval request carries the origin and the grantable classes, and arrives with neither when the effect is critical, carries a class outside `activation`/`form_mutation`, or has a destination leaving the origin. A panel that worked this out for itself would be a second copy of the rule. Six policy tests pin it: never critical, never another origin, never a class the grant does not name, never an origin the run is not already allowed on.

**The grant rides the transition the approval already causes.** A run has no status-preserving write, and inventing one to record a convenience would put a second way to move a run outside its state machine — so the scope travels with the answer and lands on the write out of `awaiting_approval`. That also means the step being approved runs under the grant, rather than the user being asked again for the effect they just widened.

**`ask_user` becomes real.** It paused with reason `user`, which is indistinguishable from the user pausing the run: the question went nowhere and nothing could answer it. The question is durable, the pause names itself (`pauseReason: "question"`), and the answer is recorded on the `paused → observing` transition that resumes the run — one commit, so a worker lost between them cannot leave an answered question still waiting. A stale panel cannot answer the question that replaced the one it showed.

Two new `@critical` scenarios: `form-prep` (three fields, one prompt, zero further prompts, nothing submitted) and `ambiguous` (asks, is answered, completes).

**Known limit, left for a decision rather than relaxed here:** a grant cannot reach a classic `<form>`. A text input inside one is `maySubmit`, which policy raises to critical, and critical is never grantable — so typing into a form field is treated as submitting it. That is policy reading "this control could submit on Enter" as "this effect submits". Relaxing it loosens an existing gate, so the fixture uses unassociated inputs and the rule is untouched.

Gate: 4 114 unit tests, 12/12 agent e2e, nine locales plus `generate:resources`. Background budget 229 500 → 231 000, measured 230 073.

Note: #372 also moves that budget line, so whichever merges second needs a trivial rebase there.

<h3>Greptile Summary</h3>

This PR adds run-and-origin-scoped approval grants and durable user-question handling, then strengthens the agent observation path following the previous review.
- Policy determines which effect classes and origins may receive a run-scoped grant.
- Question answers are recorded atomically with the transition back to observation, and ordinary resume no longer bypasses an open question.
- Recent changes make question identifiers monotonic, project observations into a smaller model-facing shape, expose full-document text and modal ownership, and size Ollama context windows from prompt length.
- Both previous findings were manually resolved after the current code added the required resume guard and monotonic question identifiers.

The PR appears safe to merge with no outstanding actionable findings.

The current code binds widened approvals to policy-provided origins and effect classes, prevents ordinary resume from bypassing questions, and uses observation-count-based question IDs that do not repeat when retained answer history is capped. The two previous threads were manually resolved without explanatory replies.

| Filename | Overview |
|----------|----------|
| packages/agent-runtime/src/controller.ts | Adds durable question answering, monotonic question IDs, resume protection, and grant persistence through state-machine transitions. |
| packages/agent-runtime/src/policy.ts | Restricts run-scoped grants to policy-approved origins and grantable non-critical effect classes. |
| packages/contracts/src/agent.ts | Defines durable grant, question, answer, and approval-offer contracts with bounded schemas. |
| src/lib/browser-agent/observation-builder.ts | Adds document-wide text collection and structural modal or landmark ownership to observations. |
| src/application/agent/agent-observation-projection.ts | Projects executor observations into a smaller model-facing representation while preserving actionable values. |
| src/application/agent/agent-model-port.ts | Sends projected observations and requests a bounded prompt-dependent Ollama context window. |
| src/features/agent/agent-view.tsx | Renders approval and question cards and disables ordinary resume while a question is open. |

</details>

```mermaid
flowchart TD
  Observe[Capture page observation] --> Project[Project model-facing fields]
  Project --> Decide[Model decision]
  Decide -->|ask_user| Question[Persist question and pause]
  Question --> Answer[Record answer and transition to observing]
  Answer --> Observe
  Decide -->|effect| Policy[Evaluate effect policy]
  Policy -->|approval required| Approval[Request user approval]
  Approval -->|once| Execute[Execute approved step]
  Approval -->|run origin| Grant[Persist offered origin/effect grant]
  Grant --> Execute
  Policy -->|covered by grant| Execute
  Execute --> Observe
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(agent): make a question the only way..."](https://github.com/shishir435/ollama-client/commit/289349e33781541f239ee60805b1d7d517fcb491) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61874645)</sub>


Shishir435:
Fixed in 289349e3, rebased onto ef25eed0 so the conflict is gone.

**Conflict.** `check-bundle-size.ts` — both PRs moved the same line. Kept both notes, since they are cumulative, then re-measured on the rebased tree: 230 700, not the number I had written while resolving. Max is 231 500.

**Resume bypass.** Correct and serious. `resume()` checked `unresolved_effect` but not `question`, so Resume walked the run past its own question to another observation with the question still attached, unanswered. `answerQuestion` is the only path that clears a question, so it is now the only path that may continue one — refused in the runtime and not offered in the panel.

**Question ids.** Also correct. From `answers.length + 1`, capped at ten while a run observes up to twenty-five times, every question after the tenth was `q11` — and since a stale answer is refused on that id alone, an old panel could answer a question it never saw. Keyed on `observationCount` now: it only goes up, and the run must observe between two questions.

4 143 tests, 13/13 agent e2e.

PR #374, opened 2026-09-09
test(agent): prove recovery against the database, and freeze the benchmark

Shishir435:
Startup recovery was asserted against a fully mocked repository — two tests — which proves the function calls the helpers it means to and nothing about whether the SQL underneath moves the run. Every transition it makes is a compare-and-set against the status the row actually holds, so a predecessor list that disagrees with the recovery path typechecks and fails only in a browser.

**11 tests now drive the real engine** at every phase a worker can be lost in:
- a cancellation settles without reissuing anything
- an effect interrupted mid-execute or mid-verify is marked uncertain **exactly once** and the run pauses unresolved, because repeating an action that may already have happened is how one click becomes two
- a read-only phase pauses with its receipts untouched
- a settled run keeps its record
- recovery is idempotent, since a restart can be interrupted too

The fixture claims its way to each phase rather than writing a status, so every hop is the transition the controller uses — which is how it surfaced that claiming `executing` refuses a run with no durable planned step.

**Benchmark** under `e2e/chromium/benchmark/`: one attempt per task family, counts, labels and hashes only so a report can be attached to an issue. No rate is published — a handful of attempts cannot support one, and the section 01 thresholds are meant to come from a clean pass rather than be guessed before one exists.

It leaked into the gate on the first try, matching `agent-*.spec.ts` *and* inheriting the harness's `@critical` tag; `--list` showed six extra entries. Both causes fixed, so the gate is 25 tests with none of them a benchmark.

Plus the Agent Preview guide: what it does, what it asks about, what it will not do, and where it stops.

**My plan was wrong about one thing.** It called for killing the MV3 worker from a Playwright spec, but this repo's own `verify-sw-turn-recovery.ts` documents that Playwright pins extension service workers alive — which is why it launches Chromium directly and kills through DevTools. A real terminated-worker agent run therefore needs its own dev-only agent-verify entrypoint, the same shape as `persistence-verify.html`. That is noted in the test's docstring and left as a separate PR rather than faked here.

`BROWSER_AGENT_AUDIT*.md` were never committed, so there is nothing to retire.

Gate: 4 154 unit tests, 13/13 agent gate, 3/3 benchmark, docs build 93 pages.

<h3>Greptile Summary</h3>

This PR adds real-database coverage for Agent startup recovery, introduces a frozen three-family browser benchmark, and documents the Agent Preview’s capabilities and boundaries.
- Exercises recovery transitions and idempotency through the real SQLite engine.
- Flushes legacy database owners after each recovery test to prevent delayed persistence from leaking across tests.
- Separates benchmark scenarios from the critical gate and records comparable, content-free metrics.
- Correctly describes site-based tab boundaries rather than promising an unenforced tab allowlist.
- Adds the Agent Preview guide to the documentation navigation.

The PR appears safe to merge; no actionable correctness, security, or repository-rule violations remain.

All previous findings are resolved in the current code, and the changes since the previous review address cleanup, retry recording, target ambiguity, backend labeling, and tab-boundary documentation without introducing a new failure.

| Filename | Overview |
|----------|----------|
| src/background/agent/__tests__/agent-recovery.smoke.test.ts | Adds comprehensive real-engine recovery coverage and flushes every test-owned legacy engine during cleanup. |
| e2e/chromium/benchmark/__tests__/benchmark-agent.spec.ts | Defines three non-gating benchmark scenarios with backend labels derived from the actual execution mode. |
| e2e/chromium/benchmark/agent-benchmark.ts | Records privacy-safe benchmark metrics, replaces retry records by scenario, and distinguishes repeated from ambiguous target identities. |
| e2e/chromium/fixtures/agent-scenario.ts | Allows benchmark scenarios to opt out of the critical-test tag without changing existing gated scenarios. |
| playwright.config.ts | Adds an isolated Playwright project for the Agent benchmark. |
| docs/src/content/docs/guides/agent-preview.md | Documents Agent behavior, approvals, limitations, and the actual site-based tab boundary. |

</details>

```mermaid
flowchart TD
    A[Interrupted Agent run] --> B{Persisted phase}
    B -->|cancelling| C[Finalize as cancelled]
    B -->|executing or verifying| D[Append one uncertain receipt]
    D --> E[Pause as unresolved_effect]
    B -->|observing, deciding, approval, takeover, pause requested| F[Pause with phase-specific reason]
    B -->|already settled| G[Leave record unchanged]
    C --> H[Second recovery is a no-op]
    E --> H
    F --> H
    G --> H
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(agent): measure what the receipts ca..."](https://github.com/shishir435/ollama-client/commit/0ffa84d515a9412ea61204399a4b012a7bc07150) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61997585)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,17 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   approvals, takeovers, and safe error vocabulary, while a new pure runtime
   package enforces legal run transitions and strict isolation from Chat. This
   foundation adds no browser permission, browser mutation, or visible Agent UI.
+- Agent Preview now completes ordinary tasks rather than only passing fixtures.
+  A page whose markup the observation could not read is observed; a wrong
+  command costs one retry with the reason instead of ending the run; each
+  decision sees what the run already did and why it turned out that way; the
+  model receives a page rather than the executor's bookkeeping, with the
+  context window sized to fit it; and an approval can be widened to a site for
+  the rest of one run, while a question the model asks is durable and
+  answerable. Submissions, destructive actions, payments, sign-ins and
+  sensitive fields keep asking every time. Startup recovery is now proven
+  against the real database at every phase a worker can be lost in, and a
+  frozen benchmark records what each task family actually costs.
 
 ## [0.13.3]
 
--- a/docs/src/content/docs/guides/agent-preview.md
+++ b/docs/src/content/docs/guides/agent-preview.md
@@ -53,8 +53,11 @@ card number or a one-time code, and it will not choose a file for you.
 
 ## What it will not do
 
-- **Only the tab you started it on.** It can open and switch to a tab, but it
-  will not wander into tabs you did not involve.
+- **One tab at a time.** Agent drives the tab it currently controls. It can
+  open a tab and switch to one, and every tab it acts on has to be readable
+  and on a site you allowed — an unfamiliar site is a new approval. It is not
+  restricted to tabs it opened itself, so treat "sites you allowed" rather
+  than "the tab you started on" as the boundary that holds.
 - **Only sites you allowed.** A destination on a new site is a new decision.
 - **Only what the page rendered.** Destinations come from links the page
   actually showed; a URL the model composed carrying data from your page is
@@ -65,6 +68,9 @@ card number or a one-time code, and it will not choose a file for you.
 ## Where it stops
 
 These are current limits, not design decisions.
+
+- **No tab allowlist.** Switching tabs is bounded by readability and the site
+  allowlist, not by which tabs the run has been involved with.
 
 - **One frame.** Content inside an iframe or a shadow root is not observed, so
   a control inside one cannot be used.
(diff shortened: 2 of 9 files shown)


Shishir435:
Fixed in 0ffa84d5.

**Guide overstated tab isolation.** Correct, and the worst of the five: nothing enforces it. `switch_tab` checks that the tab exists, is readable and its origin is allowed — so on an already-allowed origin it would not even prompt. The guide now says the boundary that holds is the site allowlist rather than the tab you started on, and "no tab allowlist" is listed among the limits.

**Distinct controls counted as duplicates.** Also correct, and my own `modal` fixture is the counterexample: two Delete buttons, one in the dialog and one behind it. Keyed on role and page as well, renamed to `repeatedTargets`, and `ambiguousTargets` is reported beside it so a number reads as suspicion rather than proof. Telling those two apart needs the owning group in the receipt, which it does not carry yet.

**Backend label.** Derived from `AGENT_HOSTED_MODEL` now, so a fixture pass cannot be labelled hosted. The fresh report reads `backend: "fixture"`.

**Retries.** A retried attempt replaces the abandoned one for its scenario instead of joining it, and the count is checked before the report is written rather than after.

**Engine flush.** Engines retained and flushed in teardown — the hazard `chat-durability.smoke.test.ts` already documents, including that coverage runs make the timer fire.

4 154 tests, 13/13 agent gate, 3/3 benchmark, docs 93 pages.

PR #382, opened 2026-09-10
feat(agent): dialogs and action-specific approvals

Shishir435:
PR 8 of the 0.14.0 agent plan.

**Dialogs.** Enabling `Page` in PR 1 made the browser stop showing `alert`/`confirm`/`prompt`/`beforeunload` — the dialog was held by the protocol with nothing able to answer it, freezing the tab. The session manager now records the held dialog with an identity, `release` dismisses whatever is still held before detaching, and a new `handle_dialog` command answers the prompt it was decided against. A blocked page is observed as blocked — tab, dialog, root frame `unreadable` — rather than asking a frozen document; controller and verifier share that seam. Every other command is refused while a dialog is open. Dismissing and closing an alert are free; accepting a confirm, prompt or beforeunload is critical and never grantable.

**Approvals.** `maySubmit` no longer escalates on its own — it is true of every field on a submit path, so typing into a search box cost an ungrantable prompt each time. Submission is priced from the resolver's own `submission` class; typing is a grantable form mutation. An edit whose target belongs to no form now says so in the approval, so the user grants against "saved as you type" rather than an implied submit step.

Also: a `beforeunload` holding the run's own navigation verifies as negative (nothing committed) instead of pausing the run, and visual clicks are covered by tests proving they inherit the semantics of the control under the point.

`typecheck`, `lint:check`, `test:run` (4503) and both bundle checks pass. No live-browser run yet; evidence is unit-level.

<h3>Greptile Summary</h3>

This PR adds end-to-end handling for native browser dialogs and refines action-specific approval pricing.
- Records blocked dialogs with stable identities and originating-document provenance.
- Adds grounded `handle_dialog` commands across parsing, resolution, execution, verification, and history.
- Prevents ordinary page commands while a native dialog is open.
- Withholds dialog text originating from unauthorized child frames.
- Prices actual submission effects separately from ordinary form mutations.
- Adds comprehensive unit coverage for dialog lifecycle, policy, authorization, visual grounding, and edit behavior.

<h3>Confidence Score: 4/5</h3>

The PR is not yet safe to merge because an unresolved authorization-boundary issue still exposes dialog text from an unapproved top-level origin.

The current adapter deliberately returns a top-level dialog’s message when the tab has redirected to an origin absent from `allowedOrigins`; the new policy requires approval before answering that dialog, but it does not prevent the model from reading the dialog content first, so the existing authorization finding remains unresolved.

**Files Needing Attention:** src/background/agent/agent-browser-adapters.ts

| Filename | Overview |
|----------|----------|
| src/background/agent/agent-browser-session-manager.ts | Tracks native-dialog identity and provenance, dismisses held dialogs during release, and coordinates dialog lifecycle with debugger sessions. |
| src/background/agent/agent-browser-adapters.ts | Produces blocked observations and protects unauthorized child-frame dialog content, but the prior unresolved concern about exposing dialog text from an unapproved top-level origin remains. |
| packages/agent-runtime/src/policy.ts | Adds action-specific dialog policy and prices explicit submissions separately from ordinary form mutations. |
| src/lib/browser-agent/resolved-effect.ts | Resolves dialog answers and attaches submission semantics only to commands that explicitly submit. |
| src/lib/browser-agent/command-executor.ts | Executes identity-checked dialog answers through the dialog action family. |
| src/lib/browser-agent/effect-verifier.ts | Verifies dialog effects through blocked observations and handles beforeunload navigation outcomes. |

</details>

```mermaid
sequenceDiagram
  participant Page as Page or child frame
  participant Session as Browser session manager
  participant Observe as Observation adapter
  participant Model as Agent model
  participant Policy as Policy evaluator
  participant Execute as Command executor
  Page->>Session: javascriptDialogOpening
  Session->>Session: Record dialog identity and origin
  Observe->>Model: Blocked observation with dialog metadata
  Model->>Policy: handle_dialog(dialogId, accept)
  Policy->>Policy: Price action against originating origin
  Policy->>Execute: Allow or require approval
  Execute->>Session: Answer matching held dialog
  Session-->>Page: Accept or dismiss
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(agent): price a dialog against the o..."](https://github.com/shishir435/ollama-client/commit/78f141f9452d76bcda866268a44b59a038bc9557) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62614339)</sub>


Shishir435:
Both findings were real. Fixed in `10b99bfb`.

**Dialog frame authorization.** The provenance was there and I discarded it — `Page.javascriptDialogOpening` names the document that opened the dialog. It now travels as `AgentDialogState.origin`; a document with none of its own (`about:blank`, `srcdoc`, an unreadable URL) reads as `"null"`, which no allowlist matches. The observation adapter withholds `message` and `defaultPrompt` for an origin outside the run's allowlist and marks it `unauthorizedOrigin` — the same treatment an unauthorized frame's elements already get, and the run is still told the dialog exists so it can ask rather than conclude nothing is there. The resolver carries that origin as `frameOrigin`, so `actingOrigin` and `baselineRisk` judge the answer against the site that asked: dismissing a frame's dialog from outside the allowlist now costs an approval naming that origin and saying its text was withheld, where dismissing the page's own still costs nothing, and no grant is ever offered for it.

**Edit persistence.** Right, and the name was the source of it — value verification compares the control and nothing else, so persistence was never established. Renamed `persistsOnChange` → `noSubmitStep`, which is what the observation actually proves, and the approval now reads "No submit step follows it, so on a page that saves as you type the change may already be stored." A test asserts the sentence does not claim the change *is* saved.

`typecheck`, `lint:check`, `test:run` (4515) and both bundle checks pass.

Shishir435:
Checked this one against the code rather than taking it as stated — the mechanism is right, the conclusion isn't, and probing it turned up a real gap next to it. Both handled in `78f141f9`.

**Why the text is not gated on the allowlist.** `observeRoot` (`agent-control-sessions.ts`) takes no `allowedOrigins`; only `observeChild` calls `authorizeAgentFrame`. So after that same unapproved redirect the run already reads the new page's every element and its entire `visibleText` — `allowedOrigins` governs child frames and where the run may travel and act, not what the document in front of it may say. Gating an `alert` there would be a stricter rule for the box on top than for the page under it, and it would blind the run to a legitimate dialog after any redirect it did not itself approve. That rule was implicit in an `origin === pageOrigin` equality check, which is a fair thing to flag; it is now stated, with the reason, in the function and in AGENTS.md.

**The real gap the probe found.** Reading a dialog and answering one are different questions, and I had only priced the frame case. `baselineRisk` escalated on `frameOrigin` outside the allowlist, so a *top-level* dialog on an unapproved origin had no `frameOrigin` and dismissal came out `allow` — free, no prompt — while a click on that same page is `activation` → high → approval. A page that navigates itself somewhere the run never approved could therefore have its dialogs answered for nothing. `handle_dialog` is now raised to high on any acting origin outside the allowlist, root frame included, and the approval names that origin instead of calling it "the page".

Three tests cover it, including one asserting the approval does *not* claim the text was withheld in the top-level case, since it wasn't.

`typecheck`, `lint:check`, `test:run` (4519) and both bundle checks pass.

PR #383, opened 2026-09-10
feat(agent): task outcomes and recovery

Shishir435:
PR 9 of the 0.14.0 agent plan.

**Three layers, kept apart.** `inputDelivery` says the page received the events; the verifier says the control changed as the step intended; neither says the goal is met. `complete` was taken at the model's word, so a run that clicked Save reported the document saved. A run that changed anything must now cite a phrase the observation actually shows, matched by the same reader `wait` uses — so a run cannot complete on evidence its own wait would reject. A run that only read owes none. Changes are counted from the resolved effect's classes and recorded durably on the receipt, because a restart keeps receipts and loses everything else; receipts that cannot be read are an unknown, not an empty history. A refusal is a safe failure: recorded as a rejected step, fed back through the run's own history, and bounded by the existing no-progress budget.

**Bounded waiting.** `wait` re-observes until the page shows its condition or the named timeout is spent, capped at six looks. It used to sleep the whole timeout and read once — the worst of both, since a save that landed in 300ms still cost thirty seconds and one that landed just after the read was reported absent.

**Recovery, proved against a real dead worker.** `pnpm verify:sw-agent-recovery` leaves a run durably `executing`, kills the MV3 worker through DevTools while the extension page and offscreen SQLite owner survive, and requires the replacement worker's own startup recovery to settle it: step `uncertain`, run `paused` for an unresolved effect, no second step. Five gates, all passing locally. Its launch and termination mechanics are now shared with the turn-recovery runner, which still passes unchanged.

Writing that gate found two real invariants I had to seed around rather than past: a run may only be created `submitted`, and entering `executing` copies the planned receipt into an execution claim in the same transaction.

`typecheck`, `lint:check`, `test:run` (4550), both bundle checks, and both worker-recovery runners pass. Bundle baselines recorded; Chrome's uncompressed total crossed 9.5MB.

<h3>Greptile Summary</h3>

This PR adds durable agent outcome evidence, bounded polling, and service-worker recovery verification.
- Requires page-supported evidence before completing runs that changed page state.
- Preserves mutation and verification information through durable step receipts.
- Adds a legal completion-refusal transition back to observation.
- Adds an end-to-end recovery gate that terminates and replaces a real MV3 worker.
- Updates critical browser scenarios to provide completion evidence.

The PR appears safe to merge; the previous correctness findings are fixed, and the current repository guide explicitly permits the recovery harness’s external DevTools usage.

Completion evidence now rejects missing, absent, self-referential, stale, and unverified claims; failed attempts no longer replace the successful mutation baseline; bounded waiting covers the full timeout; and no actionable regression remains. The raw-CDP thread is no longer outstanding because the authoritative guide now expressly excludes external `tools/verify/**` recovery runners from the shipped-extension ownership restriction.

| Filename | Overview |
|----------|----------|
| packages/agent-runtime/src/completion.ts | Introduces completion judgment using durable mutation receipts, verification state, current observation evidence, and best-effort stale-evidence detection. |
| packages/agent-runtime/src/controller.ts | Integrates completion refusal, durable rejected-step recording, mutation baselines, and re-observation into the run lifecycle. |
| packages/agent-runtime/src/state.ts | Allows a declined completion decision to transition safely from deciding back to observing. |
| src/lib/browser-agent/effect-verifier.ts | Implements bounded repeated observation while covering the complete requested wait interval. |
| tools/verify/verify-sw-agent-recovery.ts | Adds an end-to-end gate proving startup recovery after terminating a real service worker. |
| tools/verify/lib/chromium-extension-harness.ts | Shares external Chromium DevTools lifecycle operations between worker-recovery verification runners. |
| e2e/chromium/critical/__tests__/agent-outcome.spec.ts | Verifies that a confirmed Save activation cannot complete the run until the page displays outcome evidence. |

</details>

```mermaid
flowchart TD
  O[Observe page] --> D[Model decision]
  D --> C[Execute command]
  C --> V[Verify control effect]
  V --> O
  D --> J[Judge completion]
  J -->|Evidence accepted| X[Completed]
  J -->|Evidence refused| R[Record rejected step]
  R --> O
  V -->|Worker terminated mid-effect| U[Mark step uncertain]
  U --> P[Pause unresolved run]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(agent): let a declined decision obse..."](https://github.com/shishir435/ollama-client/commit/d5869bfe503fdff42539df1948e783571e383feb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62642586)</sub>


Shishir435:
Both real. Fixed in `cdf70f99`.

**Wait ended an interval short.** Confirmed by arithmetic: six looks leave five gaps, each ~timeout/6, so a 30s wait took its last look at 25s and reported absent — and the `AGENT_WAIT_MAX_INTERVAL_MS` clamp meant simply widening the gaps could not close it either. The look before the last now waits out whatever remains, so the final observation lands on the deadline: looks at 0, 5s, 10s, 15s, 20s, 30s. A test asserts that cadence and that a condition arriving in the last stretch is confirmed.

**Incidental text proving completion.** Right, and the Save-button example is exactly the bypass. One correction to the framing though: the trusted layer cannot check that a phrase *demonstrates* the goal or *relates to* the mutation — that is the claim the model is making, and no deterministic rule decides it. What it can check is the other half: evidence that was already true before the change cannot be evidence of the change. Two refusals now:

- `self_evidence` — the evidence is the acted-on control's own label, from the step's durable target. Compared exactly after normalising rather than by containment, so a goal worded around a button's text ("renamed to Save the world") is still answerable; containment would refuse more real answers than it catches bypasses.
- `stale_evidence` — the evidence was already in the page as it read when the change was decided. The controller keeps that one page's flattened text for the active run and hands it to the judge.

The baseline is best-effort by construction: it lives in the worker that made the change, so a restart loses it and the check is skipped. An absent baseline is treated as unknown, never as proof the evidence is new — the same rule the receipts branch already follows. The tool schema and system prompt now state the requirement, so a model is told the rule rather than discovering it through refusals.

`typecheck`, `lint:check`, `test:run` (4557) and both bundle checks pass; the two worker-recovery runners were re-run after the change.

Shishir435:
Real, and writing the test for it found a second bug next to it. Both in `fc4a4d78`.

**Baseline vs. the change that landed.** Correct: capture happened when a mutating command was *chosen*, while `lastChange` selects on what was *applied*, so the two could disagree. Promotion now happens against the status the step actually settled on, through `isAppliedAgentStepStatus` — exported from `completion.ts` and shared with the judge's own selection, so the two cannot drift again. A refused command, and one that executed and then verified negative, both leave the baseline alone.

**The one the test found.** My first attempt promoted on the `executed` append, which still failed. Reason: a step is appended once per lifecycle change, so `steps()` returns several receipts per step and my judge read all of them — a superseded `executed` receipt for a step that went on to fail is an applied change with no verification, which refuses every completion after it for the rest of the run. Receipts now collapse to the last per step before selecting, the way `buildAgentHistory` already does. That bug was in the original commit, not introduced by the fix.

**On the restart path being unresolved.** Worth being precise rather than agreeing wholesale. The durable half of the check survives a restart: `self_evidence` reads the acted-on control's label off the step receipt, so the Save-button example stays closed. What is lost is `stale_evidence`, and only for a change that had already verified `confirmed` before the worker died — an interrupted step comes back `uncertain`, which the unverified-change rule refuses before evidence is considered at all.

Making that half durable would mean persisting page text on the receipt. Receipts are read back into prompts and are deliberately kept clear of it — source URLs are stripped to origin and path, typed values are redacted — so storing a page's full text to close a narrow gap is a worse trade than the gap. The limitation is now stated in AGENTS.md rather than left implicit.

`typecheck`, `lint:check` and `test:run` (4559) pass. I have not re-run the two worker-recovery runners against this commit.

Shishir435:
The cancelled browser gate mattered more than either comment — it was hiding a bug that broke almost the whole agent suite. Fixed in `d5869bfe`, along with a reply to both findings.

**The run stranded itself after every refusal.** `claimAgentRunPhase` filters `expected` by the phase's legal predecessors before it reaches SQL, and `deciding -> observing` was not an edge — so the claim I make after declining a completion matched no row, `processCompletion` returned undefined, and the run sat in `deciding` until its budget expired. Fourteen of the twenty-one agent scenarios were failing on the branch; the unit suite was green throughout, because the controller's test double checked `expected` without applying the predecessor filter and was therefore more permissive than the database.

The edge exists now: a declined decision touched nothing, so looking again is the whole point of calling it a safe failure, and `state.test.ts` asserts it deliberately rather than having flipped silently. The double enforces the same filter, so this class cannot hide in unit tests again. Every scenario whose run changes the page now cites the evidence a real model would have to — not fixture busywork, it is what the new rule requires of anything that completes.

Verified locally this time: `chromium-agent` 21/21, the whole `@critical` gate 33/33, both worker-recovery runners, plus typecheck, lint and 4560 unit tests.

**On the raw CDP finding — declining, and fixing the wording that caused it.** The rule is about extension code: `chrome.debugger` attachments taken against the user's tab in the shipped product, kept out of model tools and out of other modules. `tools/verify/**` is a Node runner outside the extension driving Chromium's own DevTools endpoint, and its entire purpose is to kill and restart the extension's worker — routing that through the session manager would mean the thing under test owning the switch that kills it. It also predates this PR: `PageSession` was extracted verbatim from `verify-sw-turn-recovery.ts`, so the rule read that broadly would have been violated since that runner landed. Nothing there ships or is reachable by a model. AGENTS.md now says so explicitly, since the previous wording gave a reader no way to tell.

**On incidental text, third pass.** Still declining the part that asks the judge to establish relevance: it cannot, and I would rather the limit be stated than faked. What is checked is now three things — the evidence must be present, must not be the acted-on control's own label, and must not have been on the page when the change was decided. What remains is a model citing a true, novel, irrelevant string, which is a lie about relevance; no deterministic rule catches that, and every heuristic I considered (same region as the change, reject element names, require the expected value) refuses more correct answers than bypasses it catches. The residual gap and its reason are written down in AGENTS.md.

PR #384, opened 2026-09-10
feat(olc): proxy reliability for unresumed decisions

Shishir435:
PR 10 of the 0.14.0 agent plan.

**The leak.** Every chat request without trailing tool results starts a backend turn, and a turn whose tools the runtime calls is parked waiting for the result. The browser agent sends a full conversation per step and treats the tool call it gets back as that step's answer — so it never resumes. Only each turn's own `SUSPENDED_TURN_TTL_MS` (ten minutes) ended one, which meant a twenty-five-step run held twenty-five live runtime sessions at once.

**Why it is a bound and not a rule.** My first fix concluded that a request carrying no tool results meant the client had moved on, and discarded its parked turns. Two existing tests say otherwise, correctly: a client may legitimately start a fresh turn while still computing the result for one it left parked, and there is nothing in the wire that distinguishes that from abandonment. So the number is bounded instead — `MAX_PARKED_TURNS` (default 4), oldest discarded first when a fresh turn is admitted, never one whose resume request already exists. Raise it for a client that genuinely interleaves.

**Terminal paths.** `ChatRoutes.inspect()` reports the parked turns, pending calls and resume holds still held, because a response that stopped is not a session that ended — a test asserting only the body cannot tell them apart. Covered: plain answer, backend failure, unknown model, resumed turn, shutdown. Shutdown now disposes the turns it fails rather than leaving them to a timer nobody will see.

**Also verified rather than assumed:** the decision tool round trip over 13 consecutive one-shot steps; an `image_url` part reaching the backend through the route (not just the wire helper) so a vision decision cannot be silently flattened to text; the `MAX_PARKED_TURNS` precedence chain.

**Scope I did not claim.** "Context limits" amounts to the catalog publishing `context_length` for the client to size against, plus an over-long prompt surfacing as an error — both already covered; there is no proxy-side truncation to test. Cancellation keeps its existing queue and client-gone coverage. The pass criterion names OpenCode and Codex specifically, and this is verified against the fake backend that exists to exercise the core; a measured run through both real runtimes belongs to PR 11's evaluation.

`typecheck`, `lint:check`, `test:run` (4570, 352 in olc) pass.

<h3>Greptile Summary</h3>

This PR bounds abandoned tool-calling sessions while preserving legitimate interleaved resumes, adds explicit terminal-state inspection, and makes shutdown settle every held turn and pending call.

- Adds configurable `MAX_PARKED_TURNS` handling with oldest-first eviction.
- Covers unresumed decision loops, queued resumes during shutdown, terminal cleanup, configuration precedence, and image-part preservation.
- Documents parked-turn lifecycle and configuration behavior.

The PR appears safe to merge; the prior zero-cap behavior is corrected and shutdown now settles resume-held turns as required.

No actionable new failure remains. The zero-cap previous finding is fully fixed by enforcing a minimum of one, and the shutdown implementation now covers parked turns, resume holds, and pending-call owners. The earlier shutdown thread was manually resolved without explanation.

| Filename | Overview |
|----------|----------|
| packages/olc/src/core/chat-route.ts | Adds serialized parked-turn eviction, route-state inspection, and comprehensive shutdown cleanup across parked, resume-held, and pending-call-owned turns. |
| packages/olc/src/core/pending-tool-calls.ts | Exposes unique pending-call turn IDs so shutdown can settle turns absent from the parked map. |
| packages/olc/src/config.ts | Adds the parked-turn cap and clamps it to the documented minimum of one. |
| packages/olc/src/core/__tests__/chat-route.test.ts | Adds coverage for bounded unresumed decisions, eviction order, shutdown lifecycle, terminal cleanup, and multimodal message forwarding. |
| packages/olc/src/__tests__/config.test.ts | Verifies cap defaults, precedence, negative-input fallback, and zero-value clamping. |
| packages/olc/README.md | Documents parked-turn semantics, limits, precedence, and the minimum supported cap. |

</details>

```mermaid
flowchart TD
  A[Chat completion request] --> B{Trailing tool results?}
  B -->|Yes, live owner| C[Hold parked turn]
  C --> D[Enter single-flight queue]
  D --> E[Re-resolve owning turn]
  E --> F[Resume backend turn]
  F --> G{Calls another tool?}
  G -->|Yes| H[Park turn and release queue]
  G -->|No| I[Dispose completed turn]
  B -->|No| J[Enter single-flight queue]
  J --> K[Reap oldest excess parked turns]
  K --> L[Start fresh backend turn]
  L --> G
  M[Shutdown] --> N[Union parked turns, resume holds, and pending-call owners]
  N --> O[Fail calls, abort, and dispose turns]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(olc): settle resume-held turns at sh..."](https://github.com/shishir435/ollama-client/commit/049ba79c19935de80a4dd65d9b73d336a9d8b27a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62759499)</sub>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -336,7 +336,10 @@ In agent mode it serves a local agent runtime over `/v1/chat/completions`, so th
   and a parked call is a promise something awaits. `ChatRoutes.inspect()`
   reports what is still held so a test can tell the two apart, and shutdown
   disposes the turns it fails rather than leaving them to a timer nobody will
-  see.
+  see. It settles the union of parked turns, resume holds and whatever the
+  call registry still names — a resuming turn is deliberately taken *out* of
+  the parked map while its deadlines are suspended, so walking that map alone
+  left a live session whose calls had no timer left to settle them.
 - **A tool result belongs to one turn, or to none.** The parked-call registry is process-wide, so a follow-up releases only the calls the turn it resumes actually owns. A follow-up whose results name no live turn is refused with `400 StaleToolResults`; starting a fresh turn instead drops the result the client just produced and lets the model redo the work behind its back. The correlation is resolved twice — once to answer fast, once inside the queue slot — because a request can wait there for as long as another turn may run, and **both** of the turn's deadlines — the turn-level one and the shorter per-call one in the registry — are suspended for as long as its own resume is waiting. They ask the same question, so a fix that suspends one and not the other only moves which timer loses the result.
 - **One turn at a time is an invariant, not a hint.** A request past its deadline is cancelled through an `AbortSignal` and the queue keeps holding the slot: a task still running has not left the single-flight boundary, whatever its caller was told. If it will not stop, the queue refuses requests with `503` and names it rather than starting a second turn beside it.
 - **A browser origin is refused unless it is allowed.** The proxy listens on loopback and runs an agent, so a wildcard `Access-Control-Allow-Origin` would let any page spend a turn — a missing response header does not stop a simple request. `ALLOWED_ORIGINS` defaults to the extension schemes; a request with no `Origin` is not a page and is left alone.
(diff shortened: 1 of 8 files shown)


Shishir435:
Both real. Fixed in `049ba79c`.

**Shutdown missed resume-held turns.** Correct, and the mechanism is exactly as described: `holdTurn` calls `clearParked`, so a resuming turn is deliberately *out* of the parked map while its deadlines are suspended. Walking that map alone left the session, the hold, and — worst of the three — the calls, whose timers were cleared on the way in, so nothing was ever going to settle them and whatever the backend awaited would hang for the life of the process. Shutdown now settles the union of parked turns, resume holds, and whatever the call registry still names.

Worth noting the gap predates this PR: the old shutdown had the same blind spot. What is new is my claim in AGENTS.md that every terminal path settles the session and the slot, which this contradicted, so the bar I wrote is the one that caught it. There is now a test that drives a resume into the queue behind a slow turn and asserts `inspect()` is empty afterwards; I checked it fails against the previous shutdown before keeping it.

**Zero cap.** Right that the pre-admission sweep cannot deliver zero — it runs before the turn it makes room for parks. The resolution is a clamp rather than a rejection, because zero cannot coherently mean "park nothing": parking is how a client tool call reaches the client at all, so zero would disable client tool calling rather than bound it. One means only the turn in flight may park.

One correction to the finding's premise: writing the test showed a negative value does *not* reach the calculation — the shared number reader passes over anything below zero, so `-3` keeps the default. That is right (a negative is invalid input, not a smaller cap) and the test now states it, after I initially asserted the wrong thing myself.

`typecheck`, `lint:check`, `test:run` (4572, 353 in olc) pass.

PR #386, opened 2026-09-10
feat(agent): comparative evaluation

Shishir435:
PR 11 of the 0.14.0 agent plan. Depends on #385 (the benchmark project was failing on `release/0.14.0`; that PR fixes it and puts it in CI).

**Thirty frozen tasks across ten families**, run twice — once with the debugger attached, once against a build with the `debugger` permission stripped, which is the browser Firefox gives us. The comparison is the point; neither pass is the reference.

**Each task scores itself from the page.** A `succeeded` predicate reads the page after the run, because false completion cannot be counted any other way: the thing being measured is precisely the run's own verdict being wrong. Its counterpart is counted too — goal met and never claimed — against the status the task *declared*, so a task meant to pause is not scored as having failed.

`AGENT_EVALUATION.md` carries the numbers. Headline:

- **0 false completions on both passes**, 31 attempts each.
- The measured gain is narrow and specific: `canvas-and-visual` goes 3 of 3 to 1 of 3 without a debugger, median 2.9s to 32.5s, because no debugger means no screenshot and a task that can only be done visually cannot be done. The other nine families are identical across backends — the design working, not a null result, since native input exists for gestures a page can tell apart and none of these tasks needs one.

**Three remaining failures are named, not rounded away**, and one is new and worth attention: a click whose handler calls `confirm()` blocks the renderer, so the action can neither finish nor be confirmed and the run pauses with `unresolved_effect` before it ever sees the dialog it caused. Dialogs found already open are handled (#382); one the run raises itself is not. The file-chooser case already has a receipt flag for that exact shape and a dialog needs the equivalent — a follow-up, not this PR.

**Bugs the instrument found in itself**, all fixed here: wall time measured from module load rather than per attempt, so every first attempt was reported as taking as long as everything before it; an ungated task threw instead of recording when it did not finish, which would have hidden the most interesting rows; and missed completions were scored against completion rather than against the declared status, which called a correctly-paused task a failure.

**Not measured, and said so in the file:** no live-model pass is published. `AGENT_HOSTED_MODEL`, `AGENT_HOSTED_BASE_URL` and `AGENT_BENCHMARK_ATTEMPTS` point the same suite at a real provider and repeat it, and Ollama with `qwen3.5` is available here, but publishing a live table is a separate run and I did not want to mix measured runtime numbers with unmeasured capability ones. Tokens are read from the provider's own counts, so the columns are empty on the fixture rather than zero — nothing is estimated.

Verified: both benchmark projects 31/31, typecheck, lint, 4572 unit tests.

<h3>Greptile Summary</h3>

This PR adds a comparative agent evaluation suite that runs thirty frozen tasks against native-debugger and DOM-only backends, records page-grounded outcomes, and publishes the measured results.
- Adds reusable benchmark task declarations, attempt recording, aggregation, token accounting, and report generation.
- Adds page-grounded predicates for false- and missed-completion measurement.
- Extends the Playwright matrix to compare debugger-enabled and debugger-free builds.
- Documents measured capability differences, remaining failures, and benchmark maintenance requirements.
- Fixes the hosted workflow so ungated benchmark scenarios run as a complete live-model suite.

The PR appears safe to merge; the previously unresolved hosted-run failure is fixed and no new actionable issue remains.

The hosted filter now exempts ungated benchmark scenarios, allowing the full live-model evaluation to execute and produce its report. The manually resolved scoring finding is also addressed by page-grounded predicates, including observable page state for the rewritten answer-following task.

| Filename | Overview |
|----------|----------|
| e2e/chromium/benchmark/__tests__/benchmark-agent.spec.ts | Declares the frozen comparative task suite with independent, page-grounded success predicates. |
| e2e/chromium/benchmark/benchmark-tasks.ts | Provides benchmark wrappers, attempt recording, and shared page and decision helpers. |
| e2e/chromium/benchmark/agent-benchmark.ts | Aggregates completion quality, timing, approval, repetition, and provider-token measurements into reports. |
| e2e/chromium/fixtures/agent-scenario.ts | Supports repeated benchmark attempts, per-attempt timing, live-provider runs, and ungated hosted scenarios. |
| e2e/chromium/fixtures/extension.ts | Extends the browser fixture support needed by comparative benchmark projects. |
| playwright.config.ts | Configures native-debugger and DOM-only benchmark projects. |
| AGENT_EVALUATION.md | Publishes the measured comparison and explicitly documents limitations and remaining failures. |
| AGENTS.md | Adds repository guidance for maintaining and interpreting agent benchmarks. |

</details>

```mermaid
flowchart LR
  T[30 frozen benchmark tasks] --> N[Native debugger backend]
  T --> D[DOM-only backend]
  N --> S[Page-grounded task predicates]
  D --> S
  S --> R[Attempt records]
  R --> A[Per-family aggregation]
  A --> J[JSON and Markdown reports]
  J --> E[AGENT_EVALUATION.md]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(agent): give the ambiguous-account t..."](https://github.com/shishir435/ollama-client/commit/4184aad22c9de5671026e6f970653deaf30ee18b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62789613)</sub>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -861,13 +861,20 @@ the tables by hand.
   percentage. Tasks are declared `gated: false`, which is what makes a stalled
   run a recorded row instead of a failed test; the run that did not finish is
   the most interesting result and throwing would leave it out.
-- **A task scores itself from the page, not from the run.** Every task carries
-  a `succeeded` predicate that reads the page afterwards. False completion
-  cannot be counted any other way, since the thing being measured is precisely
-  the run's verdict being wrong. Its counterpart — goal met, never claimed —
+- **A task scores itself independently of the run.** Every task carries a
+  `succeeded` predicate: for an effect it reads the page, and for a reading
+  task it requires the page to state a fact *and* the answer to carry it.
+  False completion cannot be counted any other way, since the thing being
+  measured is precisely the run's verdict being wrong — and
+  `Boolean(run.result)` is not a scorer at all, because `result` is the
+  model's own summary and exists whenever a completion was accepted. Its counterpart — goal met, never claimed —
   is counted too, against the status the task *declared*, because some tasks
   are meant to pause and scoring those as missed would call the right answer a
   failure.
+- **A live pass runs the whole suite.** The critical suite's hosted matrix
+  deliberately runs only a couple of its tasks, and applying that skip to the
+  benchmark left a hosted run recording nothing and writing no report — the
+  opposite of the point. The skip applies to gated scenarios only.
 - **CI runs the fixture pass.** Not as a threshold gate, which it is not, but
   because it asserts, and the completion-evidence rule broke two of its
   scenarios while nothing was running it.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -848,6 +848,30 @@ Branch promotion has three stages: `release/*` → `preview` → `main`. Merge a
 - `tools/README.md` documents command ownership and prerequisites. `check:static` is shared by CI and local verification; `verify:ci-parity` runs static checks and coverage on committed HEAD in a clean worktree.
 - Never bypass with `--no-verify`. If a hook fails, fix the cause.
 
+## Measured agent behaviour
+
+`AGENT_EVALUATION.md` holds the published numbers and the named remaining
+failures; regenerate it from the two benchmark projects rather than editing
+the tables by hand.
+
+- **The benchmark records, the gates assert.** `chromium-agent-benchmark` and
+  `chromium-agent-benchmark-dom` run the same thirty frozen tasks with and
+  without the `debugger` permission — the second is the browser Firefox gives
+  us — and write counts, never rates: a handful of attempts cannot support a
+  percentage. Tasks are declared `gated: false`, which is what makes a stalled
+  run a recorded row instead of a failed test; the run that did not finish is
+  the most interesting result and throwing would leave it out.
+- **A task scores itself from the page, not from the run.** Every task carries
+  a `succeeded` predicate that reads the page afterwards. False completion
+  cannot be counted any other way, since the thing being measured is precisely
+  the run's verdict being wrong. Its counterpart — goal met, never claimed —
+  is counted too, against the status the task *declared*, because some tasks
+  are meant to pause and scoring those as missed would call the right answer a
+  failure.
+- **CI runs the fixture pass.** Not as a threshold gate, which it is not, but
+  because it asserts, and the completion-evidence rule broke two of its
+  scenarios while nothing was running it.
+
 ## Constraints
 
 - MV3 CSP blocks dynamic eval; WASM is allowed via `'wasm-unsafe-eval'`. ONNX Runtime is bundled, never fetched.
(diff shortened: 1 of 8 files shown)


Shishir435:
Conflict resolved and both findings fixed in `cc17b6a9`.

**The conflict.** #385 added completion evidence to the three old benchmark scenarios; this branch replaced that file with the thirty-task suite. The suite supersedes them and already carries evidence wherever a task changes the page, so it is taken whole. #385's CI steps are in another file and merged untouched — verified: `git diff origin/release/0.14.0 -- .github/workflows/ci.yml` is empty. Both projects re-run 31/31 on the merged tree.

**Scorers trusting the run's verdict.** Correct, and it is the more serious of the two: `Boolean(run.result)` is the model's own completion summary, so an accepted completion produced one by construction and nine tasks agreed with the run every time. They could not have detected a false completion, which is the one number this suite exists to produce.

Reading tasks now require both halves — the page has to state the fact and the answer has to carry it — so a drifted fixture fails rather than passing vacuously. `multi-tab` tasks check the tab that actually holds the answer. `ask-when-ambiguous` scores on the *user's* recorded answer, which is durable state rather than model-authored. The two editor tasks that checked only the edited value now check the save landed, since the goal says to save. Zero `run.result`-only scorers remain.

Re-measuring moved no number: still 0 false completions on both passes, same totals. That is worth stating precisely — the figure was right, but before this fix it was not evidence, and `AGENT_EVALUATION.md` now says so rather than quietly keeping the table.

**Hosted runs skipping the suite.** Right, and it made the workflow the evaluation documents impossible: `AGENT_HOSTED_MODEL` skipped all thirty tasks and the report writer, so a live pass recorded nothing. The skip exists because the critical suite's live matrix deliberately runs only a couple of its tasks; the benchmark is the opposite case, so it applies to gated scenarios only now. Verified end to end against `qwen3.5:latest` through Ollama — one task runs and scores, taking 3.6 minutes where the scripted model takes 2.7 seconds.

No live table is published. At that rate a full pass is hours, and the file states that as an explicit gap rather than implying the numbers cover a real model.

Shishir435:
Right, and the finding pointed at something worse than a weak scorer: that task had *no* honest scorer available. Its page had no controls, so nothing the run did could be observed, and the only things left were the answer the harness itself supplied and the model's own summary.

Fixed in `4184aad2` by fixing the task rather than the predicate. The page now has two account controls; the run asks which, and then has to click the one it was told, which the page records. Asking is only half the behaviour worth measuring — acting on the answer is the half that leaves a mark, and it is the half that can be checked. The scorer reads the page and also requires the *other* account not to have been chosen.

No scorer in the suite now rests on the run's own verdict: `grep "run?.result"` in the spec returns nothing, and the remaining use is inside `reportsFact`, where the answer is checked against a fact the page independently states.

Re-measured on both backends. Totals unchanged — 27 of 31 completed and 29 met on the native backend, 25 and 27 on the DOM one, 0 false completions either way — and `AGENT_EVALUATION.md` carries the fresh figures and the note about why the task was rewritten.

PR #387, opened 2026-09-10
fix(agent): refuse a back onto an unreadable history entry

Shishir435:
Follow-up to a failure the evaluation (#386) named as undiagnosed. Scope narrowed after review — see the second half.

**Fixed — `multi-tab/go-back`.** A tab commits `about:blank` before its first page, so pressing back at the start of a run's own history reached the destination parser as a non-HTTP URL and threw a bare `Error`. `agentResolutionFailure` has no case for that, so it fell through to `verification_failed` — a code about page effects, raised for a command that never touched the page. Any user pressing back to the start of a tab hits the same mislabel. Such an entry is refused now as an unreadable destination, in the same terms an unknown one already is.

The refusal is in the resolver and deliberately not in the history. Filtering `about:blank` out of the recorded stack is the tempting fix and it is wrong: the stack is walked by index, so dropping an entry would leave the run predicting one page while the browser went to another — an unapproved page reached through an approval given for a different one. That reasoning is now written down where someone would try it.

The evaluation's own task was also wrong: its second-step predicate matched the word "Details", which the home page carries as its link label, so it asked to go back twice. Fixed to key on the details page's status.

`multi-tab` is **3 of 3 on both backends**, up from 2, and the published table is regenerated from that run.

**Removed — the native-dialog receipt machinery.** The first version of this PR also added `AgentExecutionReceipt.dialog` and a classification branch, while stating they did not fix the case. Review was right that this should not merge, and right about why, so it is gone.

The reason is worth keeping though, and the evaluation now records it: the control port has **no timeout at all**, so the page-side call after a native click waits forever on a renderer blocked in `confirm()`, and nothing after it runs. A receipt flag written after the action is on a path that case never takes — a component with no real caller, which the repo's own rule says not to add. A real fix has to stop waiting on the page once a dialog is known to be holding it, which means the executor learning about the dialog from the debugger rather than from a document that cannot answer — the same principle the observation path already follows. That is its own change, with its own measurement.

`AGENT_EVALUATION.md` keeps `native-confirm` as an open failure with that sharper diagnosis.

Verified: typecheck, lint, 4574 unit tests, both benchmark projects 31/31.

<h3>Greptile Summary</h3>

The PR fixes back-navigation into unreadable browser-history entries and corrects the benchmark that exposed the issue.
- Rejects missing, malformed, and non-HTTP(S) history destinations with `AgentUnreadablePageError`.
- Keeps browser history entry-for-entry aligned with the browser rather than filtering `about:blank`.
- Updates the multi-tab benchmark predicate and regenerated evaluation results.
- Documents that native-dialog recovery remains unresolved and removes machinery that could not execute after a dialog blocked the renderer.

<h3>Confidence Score: 4/5</h3>

The history-navigation fix appears correct, but the existing repository-rule finding about observing a dialog-blocked page remains outstanding and must be satisfied before merging.

No new actionable defects were found in the changes since the previous review. The earlier dialog-timing finding remains outstanding: an action that opens `confirm()` still blocks the renderer during settlement before the executor can observe the dialog, so the run exhausts its budget instead of observing the page as blocked; this PR explicitly documents rather than fixes that behavior.

**Files Needing Attention:** src/lib/browser-agent/command-executor.ts

| Filename | Overview |
|----------|----------|
| src/lib/browser-agent/resolved-effect.ts | Validates history destinations before constructing an effect and maps unreadable entries to the established typed refusal. |
| src/lib/browser-agent/__tests__/read-only-actions.test.ts | Covers refusal of `about:blank` and preservation of ordinary HTTPS back navigation. |
| src/background/agent/agent-tab-history.ts | Documents why all committed entries must remain indexed exactly as the browser records them. |
| e2e/chromium/benchmark/__tests__/benchmark-agent.spec.ts | Keys the back action to details-page content so the benchmark no longer requests a second back navigation from Home. |
| AGENT_EVALUATION.md | Updates measured results and accurately records the native-dialog limitation as unresolved. |

</details>

```mermaid
flowchart TD
  A[Agent requests back or forward] --> B[Resolve indexed browser-history destination]
  B --> C{Known HTTP or HTTPS URL?}
  C -- No --> D[Throw AgentUnreadablePageError]
  C -- Yes --> E[Create approved destination]
  E --> F[Execute browser history movement]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["refactor(agent): drop the dialog receipt..."](https://github.com/shishir435/ollama-client/commit/e95ac4c7700682450d1d258bdbf7503afdcd5720) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62830443)</sub>

### DIFF
--- a/AGENT_EVALUATION.md
+++ b/AGENT_EVALUATION.md
@@ -122,16 +122,17 @@ while the browser went to another.
    renderer, so the action can neither finish nor be asked what it delivered,
    and the run pauses before it ever sees the dialog it caused.
 
-   Part of the machinery is now in place and does not fix it:
-   `AgentExecutionReceipt.dialog` records that an action raised one, and
-   `classifyVerificationOutcome` sends such a step back to be re-decided
-   rather than pausing — including at critical risk, which is where a
-   confirmation actually lives. That is necessary and insufficient, because
-   the executor does not reach the point of writing a receipt: the step runs
-   for the full 32-second budget, which is the shape of the dispatch itself
-   waiting on a renderer that is blocked. The remaining work is at that
-   level — the action has to stop waiting once a dialog is known to be
-   holding the page — and is not attempted here.
+   The diagnosis is sharper than it was, and it rules out the obvious fix.
+   Recording the dialog on the receipt after the action — the way a file
+   chooser is recorded — cannot work: the control port has no timeout, so the
+   page-side call that follows a native click waits forever on a renderer
+   blocked in `confirm()`, and nothing after it runs. The step burns its full
+   budget there. Any fix has to stop waiting on the page once a dialog is
+   known to be holding it, which means the executor learning about the dialog
+   from the debugger rather than from the document — the same principle the
+   observation path already follows, where a blocked page is observed as
+   blocked rather than asked. That is not attempted here, and no partial
+   machinery for it is shipped.
 2. **An over-claiming run can exhaust its budget instead of recovering.**
    `delayed-save/claims-before-it-lands` claims completion the moment Save is
    pressed. The claim is correctly refused every time — 0 false completions is
--- a/AGENT_EVALUATION.md
+++ b/AGENT_EVALUATION.md
@@ -74,16 +74,16 @@ model, where it does.
 | editors | 3 | 3 / 3 | 3 / 3 | 0 / 0 | 0 / 0 | 2920 / 2934 |
 | form-preparation | 3 | 3 / 3 | 3 / 3 | 0 / 0 | 0 / 0 | 2908 / 2938 |
 | frames | 3 | 2 / 2 | 3 / 3 | 0 / 0 | 0 / 0 | 2887 / 2942 |
-| multi-tab | 3 | 2 / 2 | 2 / 2 | 0 / 0 | 0 / 0 | 4444 / 4412 |
+| multi-tab | 3 | 3 / 3 | 3 / 3 | 0 / 0 | 0 / 0 | 2933 / 2954 |
 | read-and-extract | 3 | 3 / 3 | 3 / 3 | 0 / 0 | 0 / 0 | 2903 / 2691 |
 | shadow-roots | 3 | 3 / 3 | 3 / 3 | 0 / 0 | 0 / 0 | 2936 / 2918 |
 | single-action | 3 | 3 / 3 | 3 / 3 | 0 / 0 | 0 / 0 | 2911 / 2906 |
 
 Totals, including the task that writes the report:
 
-- **native backend** — 27 of 31 attempts reported completion, 29 met the
+- **native backend** — 28 of 31 attempts reported completion, 30 met the
   predicate, **0 false completions**, 1 missed.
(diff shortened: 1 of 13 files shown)


Shishir435:
Right, and the correct conclusion is the one I dodged: don't merge machinery that cannot be reached. Removed in `e95ac4c7`; the back-navigation fix stays.

Checking the claim turned up the underlying fact, which is worse than "detection runs too late": the control port has **no timeout at all**. So the page-side call after a native click does not merely finish late on a renderer blocked in `confirm()` — it never finishes, and nothing sequenced after it ever runs. A receipt flag written there is on a path this case never takes, which by the repo's own rule is a component with no real caller.

So the flag, its classification branch and their tests are gone rather than merged as scaffolding. What is kept is the knowledge: `AGENT_EVALUATION.md` now records why the file-chooser shape does not transfer, and what a real fix has to do — the executor has to learn about the dialog from the debugger rather than from a document that cannot answer, which is the same principle the observation path already follows for a blocked page. That is its own change with its own measurement, not a rider on this one.

One thing worth flagging beyond this PR: the missing port timeout is not dialog-specific. Any page that blocks its renderer — a long synchronous script, `alert`, a debugger statement — hangs an executor await indefinitely today, and only the run's own step budget ends it. I have not changed that here because it is a broad behavioural change that deserves its own measurement, but it is the more general version of this bug.

Shishir435:
Two things: a correction to something I wrote in this thread, and a disagreement about gating.

**Correction first.** I said the page-side call "never finishes". I verified that the control port has no timeout — that is real, and it is why a flag written after the settle is unreachable. I did **not** verify which await actually hangs, and the run's own record argues against my stronger claim: the attempt records two steps, two observations and two approvals asked before it gave up at the 32-second harness cap. Two steps means something returned. So "no timeout exists, therefore a post-settle flag is on a path this case may never take" is what I can support; "it never finishes" was more than I had measured, and I should not have written it that way one message after correcting myself for exactly that.

That is also why I am not fixing it by guessing again. Which await hangs, and whether the debugger dispatch resolves at all against a blocked renderer, is answerable — with an instrumented run, not by reading.

**On gating.** The finding is against code that is no longer in this PR; GitHub has marked the comment Outdated. The remaining objection is that `native-confirm` still fails, and that is true — but it fails identically on `release/0.14.0` right now. It is a pre-existing bug, found and published by #386, which is already merged and names it as an open failure. This PR does not introduce it, touch its code path, or make it worse.

What this PR does fix is a different bug, verified by measurement: backing onto a non-HTTP history entry died as `verification_failed` — a code about page effects, for a command that never touched the page — and any user pressing back to the start of a tab hit it. `multi-tab` goes 2 of 3 to 3 of 3 on both backends.

So I am asking for this to merge as it stands. Holding a verified, independent fix behind an unrelated pre-existing bug that needs its own instrumented investigation costs the fix and does not buy the investigation. I would rather do that investigation properly as its own change, with a measurement showing the task actually moves — which is the standard I failed to meet the first time I claimed this one fixed.

PR #388, opened 2026-09-10
feat(agent): release hardening — disclosure, supervision, actionable failures

Shishir435:
PR 12 of the 0.14.0 agent plan.

**Attachment disclosure.** Chromium shows its own "is debugging this browser" banner the moment a run attaches, and the panel said nothing about it. A banner with no explanation beside it is exactly what sends someone to ask a developer what their extension is doing. `AgentBrowserDisclosure` now travels on every snapshot — read from the session manager, not guessed from a user agent — and the panel states before a run starts what attaching means and when it ends. A browser with no debugger says what it therefore cannot do: synthetic input only, no screenshots, no native dialogs.

**Actionable failures.** `AgentError.message` is written in English for whoever reads a receipt, and says what happened rather than what to do. The panel now leads with `agent.failure.<code>` in the reader's language — narrow the goal, reload the page, choose a model that can call tools, take the tab over — and keeps the original beneath, because a bug report still needs the words the run used. An unrecognised code falls back to the `unknown` advice rather than to nothing.

**Supervision.** The status is the machine's word for it, so the panel also names the step in flight, using the same label the work log uses so the two cannot disagree; shows progress against the observation budget that will stop the run rather than a bare count; and counts every tab once the run has adopted more than the one it started on.

**Documentation was wrong, which matters more than any of the above.** The guide claimed no shadow roots, no vision, and that native dialogs cannot be seen or answered — three capabilities that shipped in #375-#382. It has been rewritten against what the evaluation actually measured, including a new section on what each browser allows, and the dialog gap that is genuinely still open is described as such rather than as a blanket "cannot".

Twenty new keys across all nine locales, `generate:resources` run. Three cards come out of `agent-view.tsx`, which the complexity gate was right to refuse at 32.

**Gates run for this PR:** typecheck, lint, 4582 unit tests, the full `@critical` browser suite (33/33), `verify:sw-agent-recovery` (5/5), and both bundle budgets — background gzip and total are within the ceilings recorded in earlier PRs, so no baseline bump was needed.

**Not covered, and worth naming:** upgrade-from-previous-version behaviour is not exercised by a test here, and no live-model evaluation table is published (both listed as gaps in `AGENT_EVALUATION.md`). The dialog-raised-by-a-click failure remains open with the diagnosis from #387.

<h3>Greptile Summary</h3>

This PR hardens the agent supervision experience and updates its documentation.
- Adds browser capability disclosure to panel snapshots and the pre-run UI.
- Presents localized recovery advice for failed runs while retaining diagnostic messages.
- Shows the current unfinished action, observation-budget progress, and adopted-tab count.
- Moves action labels into the localization system across all nine locales.
- Shares the observation limit between the contract, runtime controller, and panel.
- Corrects the guide’s descriptions of shadow-root, vision, dialog, and browser-specific capabilities.

The PR appears safe to merge; no actionable new defect or outstanding previous finding remains.

The stale-action issue is fixed by showing only the newest unfinished step, action labels are now localized across all nine locales, and the displayed observation ceiling now comes from the same exported constant enforced by the runtime. The first two previous threads were manually resolved without explanatory replies, and their underlying issues are also fixed in the current code.

| Filename | Overview |
|----------|----------|
| packages/contracts/src/agent.ts | Defines the shared observation limit and applies it to the run-state step bound. |
| packages/agent-runtime/src/controller.ts | Replaces the runtime’s duplicated observation-budget literal with the shared contract constant. |
| src/background/agent/agent-panel-port.ts | Derives browser disclosure from session-manager capabilities and includes it in every snapshot. |
| src/features/agent/lib/presentation.ts | Produces localized action descriptors, excludes settled steps from the current-action header, and maps failure codes to advice. |
| src/features/agent/agent-view.tsx | Integrates disclosure, localized current-action rendering, shared-budget progress, run details, and outcome cards. |
| src/features/agent/components/agent-outcome-card.tsx | Leads settled failures with localized recovery guidance while preserving the original diagnostic message. |
| docs/src/content/docs/guides/agent-preview.md | Updates browser capability and measured-limit documentation to match the current implementation. |

</details>

```mermaid
flowchart LR
  SM[Browser session manager] -->|capabilities| BG[Agent panel port]
  RT[Agent runtime controller] -->|run state and steps| BG
  CT[Contracts] -->|snapshot schema and shared observation limit| BG
  CT --> RT
  BG -->|panel snapshot| AP[Agent panel]
  AP --> AV[Agent view]
  AV --> BD[Browser disclosure]
  AV --> RD[Run details and progress]
  AV --> WL[Localized work log]
  AV --> OC[Localized failure advice]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(agent): localize action labels, hide..."](https://github.com/shishir435/ollama-client/commit/f00165f8d254215a1d6f5fa10c64439429cb7fb4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62846188)</sub>


Shishir435:
All three fixed in f00165f8.

**Stale action.** `currentAgentAction` now returns a label only while the newest command's step is `planned`, `approved`, `executing` or `executed`. A verified click stays the newest command all through the observation and decision that follow it; the header shows nothing there rather than claiming the run is still clicking. The work log keeps it, because the log is the record.

**Localisation.** `commandLabel` is gone. `agentActionLabel` returns `{ key, values }` and the two render sites translate it, so `agent.action.*` — 30 keys — is in all nine locales. Two shapes needed a key each rather than an interpolated value: scroll direction (`scroll_up`/`down`/`left`/`right`) and the dialog answer (`handle_dialog_accept`/`dismiss`), since `"up"` and `"accept"` have no translation at the point of use. Page-derived values (URL, wait condition, query, inspect region) still go through `agentPlainText` before they travel.

**Budget constant.** `MAX_AGENT_OBSERVATIONS = 25` now lives in contracts; the loop check, the `stepCount` schema bound and the panel's progress line all read it. The panel's own copy is deleted.

Gates: typecheck, lint, 4586 unit tests, `generate:resources`, and the 33 `@critical` e2e all pass. New tests cover the settled-step case, the flattened label values, and the direction key.

PR #403, opened 2026-09-18
feat(agent): fill a form in one step, and resolve the window from the model

Shishir435:
Closes the last correctness gate and the context work in one PR.

`fill_form` sets up to twelve controls from one decision and `extract` asks up to six scoped queries in one walk of the document — a decision costs seconds and an observation costs tens of milliseconds, so ten predictable fields were ten round trips and almost no browser work. Each batched field resolves, approves and verifies exactly as the single-field command it mirrors; the batch changes who asks, never what is checked. It cannot click and therefore cannot submit, which is what makes one approval for it honest, and it stops at the first field it cannot place and reports the count rather than throwing it away.

The context budget is resolved from the model instead of two literals. `auto` takes the smallest figure the server's allocation, the weights' metadata and the catalog give, falls back to what the agent has always reached, and settles once per run so a local runner stops reloading on a `num_ctx` that moved. History, answer and page are each bounded against it. Screenshots default to `auto` rather than every step, and both knobs are in Settings → Browser with search entries and copy in nine locales.

Two defects surfaced while testing. A batch refused its own second field on `formFingerprint` — that hash covers the form's values and the batch had just changed them — so the remaining fields are re-baselined after each applied edit. And an unknown model's window first fell to 16,384 where the agent had always reached 32,768, halving the page budget for every model that reports nothing; an existing browser gate caught the run re-reading a page it had already found its control on.

4,872 unit tests, 37/37 `chromium-agent` gates, both bundle budgets raised with ledger entries (total 9,626,409; background gzip 269,969).

**Review fixes.** The batch approval now names every control it will set, not the first of twelve, and no longer promises that a change cannot be stored. `extract` asks every child frame the run may read and merges their matches into the group that asked — proving that found a second defect, since the browser adapter had been dropping `lookup` entirely, so `extract` reached the page as a plain read from the day it shipped. Under `auto`, the first step on a page the run has not seen gets a picture, which is what keeps `zoom` reachable after a navigation.

4,885 unit tests, 38/38 `chromium-agent` gates.

<h2><a href="https://app.greptile.com/api/retrigger?id=66242213"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the previous child-frame extraction issue is resolved without introducing a new actionable failure.

<h3>Summary</h3>

This PR adds batched form filling and multi-query extraction, resolves the agent context window from model and server metadata, and introduces configurable automatic screenshot capture.

Changes since the previous review repair the remaining child-frame lookup budget issue:
- Reserve bounded element capacity before observing the root document so readable child frames can answer every extraction query.
- Mark lookup results as truncated when the element budget prevents a child frame from being searched.
- Add focused tests for frame-budget reservation and truncation semantics.

All previous Greptile threads are resolved, and the current implementation addresses their reported failure modes.

<details><summary>Diagram</summary>

```mermaid
flowchart TD
  A[Extract decision] --> B[List selected child frames]
  B --> C[Reserve bounded child lookup capacity]
  C --> D[Observe root with reduced element limit]
  D --> E[Observe each authorized child]
  E --> F{Child receives budget?}
  F -->|Yes| G[Merge matches by query]
  F -->|No| H[Mark lookup groups truncated]
  G --> I[Apply per-query match limit]
  H --> I
  I --> J[Return composed observation]
```
</details>

<sub>Reviews (3) · Last reviewed commit: ["fix(agent): hold room for a frame&#39;s answ..."](https://github.com/shishir435/ollama-client/commit/5aa11558f41b1a6d89cecae23ee948b2da6f0321)</sub>


Shishir435:
Fixed all three, plus one the fix for the second uncovered.

**Batch approval hides targets** — the evidence block now lists every control the batch will set, in order, from each resolved target's own accessible name, with the role/input type/tag as the fallback for a control the page never named. Shown at the bound the request schema already allows, in a box that scrolls instead of clipping at three lines. The consequence no longer says "nothing is submitted": the batch presses nothing so it cannot submit, but a page that saves as you type has already stored each change.

**Extract skips child frames** — the lookup is asked of every child frame the run may read, and each frame's matches merge into the group that asked for them; refs carry their frame in their prefix, so one group spanning two documents stays unambiguous. The per-question bound is applied after the merge, so three frames cannot return thirty rows and call none of them truncated. A scope stays root-only — its answer is a count and a continuation offset, and neither composes. Writing the browser gate for this found a second defect: the browser adapter forwarded `scope` and dropped `lookup`, so `extract` had been reaching the page as a plain read since it shipped and every group came back empty.

**Auto vision loses recovery** — the first step on a page the run has not seen now warrants a picture, compared against the page the last recorded step acted on in the shortened form history already carries, so a query or fragment moving buys nothing. One picture per page, which is what keeps `zoom` reachable after a navigation.

4,885 unit tests, 38/38 `chromium-agent` gates (the new one is `extract-reaches-an-authorized-frame`).

Shishir435:
**Child lookup can be skipped** — fixed, and the failing shape is worse than it reads: a lookup is asked precisely when the control cannot be found, so the root matches nothing and falls back to its overview. On a crowded page that fallback filled the element budget before any frame was asked, and the frame holding the control was the one skipped.

The reserve is now taken before the root is asked — one group per question per listed frame, capped at half the element budget, withheld from the root's own request. It costs the root's consolation rows only, since a lookup's own matches are bounded far below that. Held for every listed frame rather than every authorized one, because whether a frame is authorized is not known until it is asked.

A frame the budget still stops now marks every group `truncated`, which is the difference between "nothing matched" and "we stopped looking". A frame the run may not read does not mark anything: the frame list already says so, and an advertising frame would otherwise leave the flag set on most of the web.

**Auto vision** — that thread is from the first review and the page-transition half of it is in `0d1da3b`: the first step on a page the run has not seen warrants a picture, compared against the last recorded step's source URL. The other half you suggest, signs of unrepresented rendered content, has nothing to read off today — the observation carries no canvas, video or embed signal, so that would be new perception rather than a new rule over the same evidence.

4,888 unit tests, 38/38 `chromium-agent` gates.

